PR #44, opened 2023-03-11
Warning before the multisig creation if balance is too low

### LINKED ISSUE #28: Check balance at creation
How to reproduce
- create a multisig (with proxy) with an account with not enough balance
- proxy creation fails because of not enough funds to pay for the reserved funds

### LINKED ISSUE #32: Show error when changing proxies
- Pure needs funds to add/change proxy
- no error showed currently

### LINKED ISSUE #43: Remove duplicate code for Tx fee calculation
`useGetCreationNeededFunds.tsx` is using some code that is also in the creation flow.
We should refactor this part to get it only once.

### LINKED ISSUE #50: Add balance check of signatories to "Change multisig" modal
In addition to checking the pure balance we should also consider checking the balance of the signatories on the "Change multisig" modal, here:

change

When we can already know that the attempt is going to fail the user can be prevented from signing the tx, seems a better experience than being stuck here: 

balance

Tbaut:
Thorough calculation of all the fees needed.
The creator doesn't send arbitrary funds to the new multisig, rather only what's needed.
Similarly the reserved funds needed for the creator to be able to create a multisig proposal is specifically calculated, and it should be chain agnostic and future-proof

closes #28 
closes #32 
closes #43 
closes #50

Here's a recap of the checks:
- Multisig creation
  - Signer has enough funds to send to the multisig (the multisig gets funds reserved for the pure proxy creation) 
  - Signer has enough funds for multisig proposal (reserved funds) and tx fees
- Multisig rotation
  - Pure proxy has enough funds to add a proxy (it's reserving funds depending on the amount of proxies)
  - Signer has enough funds for both multisig proposals (reserved funds) and tx fees
- Send Tx
  - Signer has enough funds for multisig proposal (reserved funds) and tx fees
  - The from address (proxy or multisig) has enough funds according to the amount to be sent


asnaith:
This is working really well and clears up a lot of confusion, nice that the UI gates the user from proceeding when necessary criteria isn't met 🙏

asnaith (approved):

PR #52, opened 2023-03-16
Better error handling on multisig rotation

### LINKED ISSUE #37: Better error handling and allow closing modal at multisig rotation
Allow closing the modal of [ui/src/components/modals/ChangeMultisig.tsx](https://github.com/ChainSafe/Multix/pull/35/files#diff-2708ec9a58ba1e1efabd56c43a36d2e0ba1ee879b1008ceac896d9a038d3009e) in `handleClose`

Tbaut:
closes #37 

Here's how you can get the error:
- initiate a multisig rotation, go through both calls, it should work.
- don't approve/reject any proposal and initiate the exact same rotation.
- if you reject the first addProxy proposal, you should be able to re-initiate the same multisig rotation, the first call should work, the second should fail with a nice error message.

### DIFF
--- a/ui/src/components/modals/ChangeMultisig.tsx
+++ b/ui/src/components/modals/ChangeMultisig.tsx
@@ -277,6 +277,7 @@ const ChangeMultisig = ({ onClose, className }: Props) => {
             isSwapSummary
             balanceMin={neededBalance}
             isBalanceError={!hasSignerEnoughFunds}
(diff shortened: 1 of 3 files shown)


asnaith:
Looking really good, thanks for the tip on how to invoke  

changes

asnaith (approved):

PR #57, opened 2023-03-21
Truncate names and fix their display

### LINKED ISSUE #31: Character overflow for multisig names
Multisig name and account name needs a character limit, also if you create a huge multisig name it breaks out of the layout on the review screen (large account name are ok because they get truncated).

Tbaut:
closes #31 

I truncate them (any new one) arbitrarily at 50 char.
I found what I did wrong in my tests, I had spaces in my names, which wrapped correctly on several lines.


asnaith (approved):
Working well 🚀

PR #63, opened 2023-03-24
Allow anyone non-propose to submit the `asMulti` if the threshold was met

### LINKED ISSUE #62: Cannot create pure when remaining approvals are signed at the same time
**Description**
If the 2 remaining signatories both sign the createPure very close to each other then the pure will not be created and there's no way to re-attempt the process from the UI. This leaves the users with a multisig only and no way to re-attempt the pure creation. 

**Prerequisites**

Use 3 accounts, 2 browsers 

**Steps To Reproduce**

Add clear steps on how to reproduce the bug

1. Connect to multix with 2 browsers
2. Create a new multisig with a threshold of 3 / 3 
(At this point the original signatory has approved the creation of the pure already and two more approvals are needed from the other multisig members to meet the threshold)
3. Open each `proxy.createPure` proposal in a different browser and click approve to invoke the Polkadot{.js} extension window (don't sign just yet)
4. Put the two Polkadot extension windows beside each other and sign one and then immediately after sign the other 

**Observed behavior**

The approvals are successful but the pure never gets created, approvals have already been completed so in the UI we cannot open a proposal again. The original proposer can cancel (but the result is a multisig without a pure) and that's the only operation that can be done in the UI. 

**Expected behavior**
After the two approvals are signed the pure should then be created.

**Screenshots / Video**

https://user-images.githubusercontent.com/3982665/227627420-ab81c3b7-f792-4344-a4cd-6e885f7adc2d.mov

**Test Host Specifics: **
 - OS: macOS
 - Browser: FireFox
 - Version 111

Tbaut:
closes #62


asnaith:
Looks good 🚀

- Re-tested with the accounts already in the predicament and I was able to recover and complete the flow
- Tested with brand new accounts, approved the `createPure` simultaneously with the two remaining approvers. One of the attempts showed a "canceled" toast and then when I approved it a second time I was able to complete the flow

asnaith (approved):

PR #77, opened 2023-04-11
Add Multix bug report template

asnaith:
Self-explanatory, thought it would be good to have one.


Tbaut (approved):
Awesome, thank you very much!

PR #90, opened 2023-04-13
Add creation time for multisig call and proxy creations

### LINKED ISSUE #36: Make sure the event handlers use an array oriented approach
Avoid the use of `getOrCreateAccount` that doesn't work with arrays

Tbaut:
closes #36

we can't upgrade the front-end yet because of a regression issue https://github.com/dotansimha/graphql-code-generator/pull/9301


asnaith:
@Tbaut Looks like that regression issue is now fixed and merged to their master but guess we have to wait for a release before updating/testing our front end?

Tbaut:
yes, I could push now, things would still work, but if we want to sync the front with the back, things would break.

Tbaut:
I want to get this merged so as not to be a bottleneck, I downgraded the package that was annoying, and we can deal with this later.

PR #91, opened 2023-04-13
Update ToastBar.tsx

Prathamesh48:
By wrapping the Snackbar component inside a Boolean(toast) check, the Snackbar will only be displayed when there is a toast object to be displayed


Tbaut (changes requested):
What problem does it solve?

Prathamesh48:
> 

the toasts will not overcome of each other

Tbaut:
This doesn't compile and doesn't make sense to me sorry.

PR #96, opened 2023-04-19
Add some documentation

### LINKED ISSUE #30: More docs around multisig creation
This needs to be investigated to be mobile friendly (not the case with tooltips)

Tbaut:
closes #30 

- sorry for the noise with the space changes, I've setup eslint ignore to hopefully not have this any more.
- adds some info when creating a multisig
- enhance the design a bit (with MUI Alerts)
- created a "Help" page to document the current setup (it's dynamic, it shows some warning too eventually, e.g when threshold is the same as the amount of signatories)
- creates a nice overview of the signatories, multisig(s) and pure proxy. Let's see if this helps someone. At least is looks kindda nice :)
- fixes a bug in the indexer for "Staking" type of proxy
- adds the proxy type in the UI for each multisig (we had the info pulled from the indexer but never used, now I use it in the diagram)

If you want to see this too, you need to
- point your local node at Kusama (see the .env.example comments in `./squid`)
- connect the UI to it using ?network=local
- add a hacky watch address such as "HqfNL8eE5Efvw5Wcj8d3tjeoCS3BPfshMHWG6rmtYUrgyrU" who's part of the ChaosDAO multisig (see the Readme of `./ui` to know how to do this)
image


Tbaut:
Merging this as I need to get going and fix #99 which will touch the same components.

PR #97, opened 2023-04-20
feat: squid dockerization

rbadillap:
_Thank you for the patience @Tbaut_

**In this PR you'll find**

**Changes in Makefile**
The changes made in the Makefile allows me to separate two important processes: 1) start the db, and 2) start the application.
Both steps needs to be separated (because in the middle we need to run `make migrate` constantly). From a deployment perspective, we need to control when we want to migrate and when we want to start the app only.

For local development purposes, you shouldn't see any differentiation, just the fact that starting from now, in order to start the app, you need to:

1. `npm ci`
2. `make codegen` and `make typegen`
3. `make build`
4. `make db` -> this is new, only start the database container
5. `make migrate`
6. `make up` -> start the app assuming the database already exists

**Created a `docker-compose.yml` file**
I'm using docker compose to facilitate the process to start the app. This is important for the people (like me) that needs to run the application in docker without any dependency in my local machine (like nodejs), since this is the most accurate scenario in production environments.

**TODO**
From this PR, I'd suggest to edit the [documentation](https://github.com/ChainSafe/Multix/tree/main/squid#quickly-running-the-sample) to indicate that `make database` needs to be executed in the middle of `make build` and `make migrate`.

**Other small changes**
1. added a root .gitignore for global files ignore definition
2. fixed Dockerfile
3. integrate processor and graphql in the `docker-compose.yaml` file that's already using squid


squid/.env.example
```diff
@@ -1,4 +1,5 @@
 DB_PORT=5432
+DB_HOST=db
```
Tbaut, line 2: What is this used for?

squid/Dockerfile
```diff
@@ -24,15 +24,14 @@ COPY --from=deps /squid/package.json .
 COPY --from=deps /squid/package-lock.json .
 COPY --from=deps /squid/node_modules node_modules
 COPY --from=builder /squid/lib lib
+COPY .env.example .env
```
Tbaut, line 27: I'm not sure we should do this, the example file can change and it shouldn't affect the deployment.

squid/docker-compose.yml
```diff
@@ -1,12 +1,43 @@
 version: "3"
 
 services:
+  node:
+    container_name: squid-node
```
Tbaut, line 5: Can we name it, graphql-server (you can add node if you wish, to remember it's a node js, but that's easy to see)

squid/docker-compose.yml
```diff
@@ -1,12 +1,43 @@
...
+      - "${GQL_PORT}:4350"
+    mem_limit: 300m
+    mem_reservation: 100m
+    command: "npm run query-node:start"
+    depends_on:
+      - db
+      - processor
+
+  processor:
+    container_name: squid-processor
```
Tbaut, line 22: Can we name it "squid-indexer" please?

squid/docker-compose.yml
```diff
@@ -1,12 +1,43 @@
...
+    build:
+      context: .
+      target: processor
+    env_file: .env.example
```
Tbaut, line 26: We shouldn't use the example here, it's not future proof.

Tbaut (commented):
Thanks a lot Ronny. I left a couple comments, mostly around naming so that things are clearer to me. There's a DB filled by and "indexer", and queriable using a "graphql server". Those keywords are important to me 🙈

Regarding the deleted files, do you know why?

squid/.env.example
rbadillap: My intention is to fully use docker to run the apps, with no need our machines to have Nodejs, NPM or any other tool installed beyond Docker.

By default, Squid attempts to connect to the database to `localhost`.

In docker environments, this is not necessary true. You can imagine docker containerization as a separate network layer from your machine, so, that means, when `docker-compose` define the services you want to start, they can be allocated in different IP addresses.

When you run: `docker compose down` and `docker compose up` the IP addresses of the services might be different every time.

In order to connect one service to the other, we use the name we set in `docker-compose.yaml`. As you can see in this Pull Request, I've added two more services, and the postgres database is called `db`, [here is the line of the code related](https://github.com/ChainSafe/Multix/blob/feature/squid-dockerization/squid/docker-compose.yml#L34).

Basically, what I am saying in:
```env
DB_HOST=db
```
Is that squid can locate the database in the service called "db", and docker will make sure to replace "db" to the proper IP address generated by them.

squid/docker-compose.yml
rbadillap: This is an interesting point.

Taking into consideration that `docker compose` will be used only for local development purposes (I'm not pretending to use docker compose to build the image while deploying a new version of the app), how would you plan to define env variables locally?

My point is, we need to define the env variables first before running `make up` as [documented here](https://github.com/ChainSafe/Multix/blob/feature/squid-dockerization/squid/README.md#quickly-running-the-sample). So my question is, how do you plan to define these env variables before starting the app? We can leverage Makefile but I'm interested to know your position.

As soon as we have clear this way, I can change the docker-compose to use a proper `env_file` or define the `env` variables manually.

rbadillap:
> Thanks a lot Ronny. I left a couple comments, mostly around naming so that things are clearer to me. There's a DB filled by and "indexer", and queriable using a "graphql server". Those keywords are important to me 🙈
> 
> Regarding the deleted files, do you know why?

@thibaut the deleted files are auto-generated while running `make codegen` and `make typegen`. I'm not deleting them from the file system, I'm deleting them from my commit since they don't contribute to my dockerization.

They are different with the generated files pushed in the repository, because mines, have different indentation, [I left a file in my commit](https://github.com/ChainSafe/Multix/pull/97/files#diff-e52ccfb35462280afdb121cca6730ad4d837df94fae3fceeda62322266007c5f) so you can take a look and determine what's might be the cause of this issue.

I believe we can define a standard using `.gitattributes` file.

rbadillap:
> Thanks a lot Ronny. I left a couple comments, mostly around naming so that things are clearer to me. There's a DB filled by and "indexer", and queriable using a "graphql server". Those keywords are important to me 🙈
> 
> Regarding the deleted files, do you know why?

@thibaut the deleted files are auto-generated while running `make codegen` and `make typegen`. I'm not deleting them from the file system, I'm deleting them from my commit since they don't contribute to my dockerization.

They are different with the generated files pushed in the repository, because mines, have different indentation, [I left a file in my commit](https://github.com/ChainSafe/Multix/pull/97/files#diff-e52ccfb35462280afdb121cca6730ad4d837df94fae3fceeda62322266007c5f) so you can take a look and determine what's might be the cause of this issue.

I believe we can define a standard using `.gitattributes` file.

squid/.env.example
Tbaut: Thanks for the explanation. But why is this a variable (I mean an environment variable) this will never change isn't it?

squid/Dockerfile
Tbaut: yup, the `.env.example` should serve as a base for e.g github secrets, and passed in deployment script the way we want.

squid/docker-compose.yml
Tbaut: well we can copy a local `.env` inside the docker, that's fine. The `.env.example` is only an example, and shouldn't be used anywhere. If the user want to use exactly these values, fine, then they copy/paste them into an `.env` that is never committed to GH, because it's their local setup.

squid/.env.example
rbadillap: This is a variable because it may depends in the way we decide to run the app in local machines, staging and production.

For example, for local environments, you may want to use a Postgres database located in our docker machine, so this value should be: `DB_HOST=db`. But, for production environments, we may want to use AWS RDS (a managed service that provides endpoints of already-served databases with no management needed), in that case the value should be: `DB_HOST=endpoint-provided-by-aws.rds.amazonaws.com`.

squid/docker-compose.yml
rbadillap: So the correct process is:

- Before running any docker command (or make command), set your environment variables first
- Run make build
- Run make up

If that's correct, then you are right, we can assume that an `.env` file should exists in the root of the squid folder and create a volume between your local machine and the docker container, so the app inside docker can access this file.

This is a good approach and we just need to improve the documentation. Let me know if you agree 👍🏼

squid/.env.example
Tbaut: got it thanks for the explanation.

squid/docker-compose.yml
Tbaut: sorry for the late reply, yup I agree.

squid/Dockerfile
```diff
@@ -7,9 +7,13 @@ FROM node-with-gyp AS builder
 WORKDIR /squid
 ADD package.json .
 ADD package-lock.json .
-RUN npm ci
 ADD tsconfig.json .
 ADD src src
+ADD typegen.json .
+ADD schema.graphql .
```
Tbaut, line 13: This is duplicated

squid/Dockerfile
```diff
@@ -7,9 +7,13 @@ FROM node-with-gyp AS builder
 WORKDIR /squid
 ADD package.json .
 ADD package-lock.json .
-RUN npm ci
 ADD tsconfig.json .
 ADD src src
+ADD typegen.json .
```
Tbaut, line 12: I wonder if this is needed or not. Types are only needed for development, once the JS is generated (with build) no need for this)

squid/Dockerfile
```diff
@@ -26,17 +30,13 @@ COPY --from=deps /squid/node_modules node_modules
 COPY --from=builder /squid/lib lib
 RUN echo -e "loglevel=silent\nupdate-notifier=false" > /squid/.npmrc
 ADD db db
-ADD assets assets
 ADD schema.graphql .
```
Tbaut, line 33: duplicated
```suggestion
```

squid/docker-compose.yml
```diff
@@ -1,15 +1,44 @@
 version: "3"
 
 services:
+  graphql-server:
+    container_name: graphql-server
+    build:
+      context: .
+      target: graphql-server
+    env_file: .env.example
```
Tbaut, line 9: This is still here, it shouldn't. if anything, then we should have the .env as environment file
```suggestion
    env_file: .env
```

squid/docker-compose.yml
rbadillap: Oh, sorry, I just made the commit to change that.

squid/Dockerfile
rbadillap: That stage of the Dockerfile called `builder` is the responsible to build the application and generate the distributable files that will be deployed to the servers.

I'd say, if `typegen.json` is necessary to build the app, let's leave it as it, but, if you say that this file is not mandatory to build the app, I can remove it.

Two lines below, I am running:
```dockerfile
RUN npm run codegen
RUN npm run typegen
```

Let me know if I should remove the typegen command there also.

squid/docker-compose.yml
```diff
@@ -1,15 +1,44 @@
...
 
 services:
+  graphql-server:
+    container_name: graphql-server
+    build:
+      context: .
+      target: graphql-server
+    env_file: .env
+    environment:
+      DB_HOST: db
```
Tbaut, line 11: if this is already in the `env_file`, why do we need it here?

squid/Makefile
```diff
@@ -26,17 +46,7 @@ generate-migration:
...
-
 typegen:
 	@npx squid-substrate-typegen typegen.json
 
-
-up:
-	@docker-compose up -d
-
-
-down:
```
Tbaut, line 38: Should we keep this?

Tbaut:
Sorry I messed up, I wanted to merge #112 into yours, and then this one into main to keep your attribution.. but I ended up merging mine into main right away. Closing this one as a result. The new dockerfile is in main. I'll make a PR to remove the makefile now.

PR #98, opened 2023-04-20
add mobile menu

### LINKED ISSUE #8: Mobile friendly menu
On Mobile, the menu doesn't look good because of the address selection.
Also we will soon add the network selection (it's already available if you uncomment the `<NetworkSelection/>`), which will clutter the menu even more.

We would need a slider for smaller resolutions, something like https://medium.com/@tsubasakondo_36683/create-responsive-drawer-menu-with-react-material-ui-617a42764b69 to place all the menu items into it.

Lykhoyda:
Closes #8


squid/.gitignore
```diff
@@ -1,3 +0,0 @@
-node_modules
```
Tbaut, line 1: Can we add this back?

squid/.vscode/settings.json
```diff
@@ -1,6 +0,0 @@
-{
```
Tbaut, line 1: And this too please

ui/src/components/Header/index.tsx
```diff
@@ -1,61 +1,66 @@
...
+  drawerWidth: number
 
-const Header = ({ className }: Props) => {
-  const { accountList } = useAccounts()
+}
+
+const AppBar = styledMaterial(MuiAppBar, {
+  shouldForwardProp: (prop) => prop !== 'open'
+})<AppBarProps>(({ theme, open, drawerWidth }) => ({
+  transition: theme.transitions.create(['margin', 'width'], {
```
Tbaut, line 26: It's magic to me, could we use good old css if that's possible? it's easier to understand and maintain.

ui/src/components/Header/index.tsx
```diff
@@ -1,61 +1,66 @@
...
-            component="div"
-            sx={{ flexGrow: 1 }}
-          >
+    <AppBar drawerWidth={drawerWidth} position="fixed"  open={open} className={className}>
+      <Toolbar>
+        <Typography
+          variant="h6"
+          noWrap
+          component="div"
+          sx={{ flexGrow: 1 }}
```
Tbaut, line 48: no `sx` please.

ui/src/components/layout/Main.tsx
```diff
@@ -0,0 +1,84 @@
...
+import {About, Creation, Home} from "../../pages";
+import {styled, useTheme} from "@mui/material/styles";
+import IconButton from "@mui/material/IconButton";
+import ChevronLeftIcon from "@mui/icons-material/ChevronLeft";
+import ChevronRightIcon from "@mui/icons-material/ChevronRight";
+import DrawerComponent from "../Drawer/Drawer";
+
+const DRAWER_WIDTH = 240;
+
+const Main = styled('main', {shouldForwardProp: (prop) => prop !== 'open'})<{
```
Tbaut, line 16: I won't comment for each, I'd like the code to stay as consistent as possible.

ui/src/components/layout/Main.tsx
```diff
@@ -0,0 +1,84 @@
...
+  justifyContent: 'flex-start',
+}));
+
+function MainLayout() {
+  const theme = useTheme();
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <Container maxWidth="lg">
+      <Box sx={{display: 'flex'}}>
```
Tbaut, line 49: same, I'll stop commenting on the sx, I'd like to avoid them at all cost.

Tbaut:
Since it's a draft, but you requested my review, I'm not sure how much I should review the actual functionality. Sorry if this particular comment isn't helpful and you knew about it:
The issue https://github.com/ChainSafe/Multix/issues/8 I wrote isn't super precise, but I meant specifically that the address selection was what was the most mobile unfriendly. This problem isn't solved:

image

Also I think we should have the menu drawer
- not visible/actionable when on a large enough screen since it adds a click and hides the menu entries. Said differently, the menu should stay the same as it is right now on destop, and the mobile menu should be visible when on a small screen
- it should not push the content, it should be above it ideally, otherwise the whole content jumps all the time

squid/.gitignore
Lykhoyda: I moved it to the root level as it's not necessary to duplicate the same configuration. Also, it's a good practice to have configuration files like eslintrc, prettierrc, nvmrc, gitignore on the root level to avoid configuring the same things in one project.

squid/.vscode/settings.json
Lykhoyda: vscode settings it's about personal preferences, similar to .ide as in my case I use a Webstorm. To ensure the same formatting styles we have a client configured. Do you know if there are any use cases I am missing to keep the vscode settings?

squid/.gitignore
Tbaut: Agreed, there are many such things not really clean in the repo, thanks for pointing that out.

we should just eventually do this in its own PR, and make sure we don't break much.

ui/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {useAccounts} from "../../contexts/AccountsContext";
+import {useMediaQuery} from "@mui/material";
+
+const routes = [
+  {url: '/', name: 'Home'},
+  {url: '/create', name: 'New Multisig'},
+  {url: '/about', name: 'About'},
+]
+
+const DrawerHeader = styled('div')(({theme}) => ({
```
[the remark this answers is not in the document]
Lykhoyda: 100% I prefer to have styled components under the main component as well. I missed it in this case

squid/.vscode/settings.json
Tbaut: I thought it was a bad idea to commit this indeed, but I've been struggling with this setting personally and wanted to make sure not to receive PRs with 100s of indentation changes. I've added an eslint rule subsequently, this is then probably not needed any more.

ui/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+  display: 'flex',
+  alignItems: 'center',
+  padding: theme.spacing(0, 1),
```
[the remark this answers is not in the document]
Lykhoyda: The framework has a predefined helper that will resolve up to 4 arguments to "spacing" method. The configured scaling factor is 8px, so it will be transpiled into:
```css
padding: 0 8px
```
f.e. in case: padding: theme.spacing(1, 2, 0, 3)
```css
padding: 8px 16px 0 24px
```
P.S. I took those codes from the examples, but we can simplify it using rem units.

ui/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+                </ListItemButton>
+              </ListItem>
+            )))}
+        </List>
+      </>
+    )
+  }
+
+  return (
+    <>
```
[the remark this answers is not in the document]
Lykhoyda: Not really, we return two Drawer components and show one depending on the mobile version or desktop

ui/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <Drawer 
+        sx={{
```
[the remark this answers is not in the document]
Lykhoyda: In general yes, we can get rid of sx, but I'm curious to know why as looks like Mui Library uses it a lot. Do you have in mind styled-components as a main library for styling?

ui/src/components/Drawer/Drawer.tsx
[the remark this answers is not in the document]
Tbaut: Sounds good, yeah I imagine it was from somewhere else. I think having padding set directly would be easier to read.

ui/src/components/Drawer/Drawer.tsx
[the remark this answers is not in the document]
Tbaut: A sorry, I read too quickly.

ui/src/components/Drawer/Drawer.tsx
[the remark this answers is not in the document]
Tbaut: This is a personal preference, being able to copy/paste CSS from the browser. And then having all the styling consistently as such in styled component rather than in sx.

In the future, we will have some components designed around the ones from MUI indeed, but we can still use styled component.

ui/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    drawerWidth: number,
+    handleDrawerClose: () => void
+}
+
+function DrawerComponent({open, handleDrawerClose, drawerWidth}: DrawerComponentProps) {
+  const { accountList } = useAccounts()
+  const theme = useTheme();
+  const matchesSmallDevices = useMediaQuery(theme.breakpoints.up('sm'));
+
+  function renderMenu() {
```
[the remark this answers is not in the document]
Tbaut: Sounds good to me.

ui/src/components/Header/index.tsx
```diff
@@ -1,78 +1,81 @@
...
+    display: none;
+    
+    @media (min-width: ${theme.breakpoints.values.sm}px) {
+        flex-grow: 1;
+        display: flex;
+    }
+`)
+
+
+const TypographyStyled = styled(Typography)`
```
Tbaut, line 67: Can use css classes here? So in the code:
```
<Typography
  className="appTitle"
  variant="h6"
  noWrap
/>
```

and then in the css at the end:
```
  .appTitle {
    flex-grow: 1;
  }
```

ui/src/components/Header/index.tsx
```diff
@@ -1,78 +1,81 @@
...
+  flex-grow: 1;
+`
+
+const IconButtonStyled = styled(IconButton)(({theme}) => `
```
Tbaut, line 71: Same comment as for the Typography component, I think we can go for all classes.
Also for this one, I noticed a glitch where for the first time you're visiting Multix, when you haven't connected your account yet, there's no menu available, so there should be no icon either. This button should be conditioned to `!!accountList?.length` and to avoid the repetition, we should go with
```js
const isShowMenu = useMemo(() => !!accountList?.length ,[accountList])
```

to avoid this
image

ui/src/components/Header/index.tsx
```diff
@@ -1,78 +1,81 @@
...
+                component={Link}
+                to={url}
+                className="buttonHeader"
+              >
+                {name}
+              </Button>
+            ))}
+          </BoxStyled>
+        )}
+        <MultiProxySelection/>
```
Tbaut, line 42: can we add the `MultiProxySelection` and the `NetworkSelection` to the mobile menu?
The Kusama deployment is imminent and while the mobile view is better without the menu, I think it won't be manageable with both the network selector and the account selector in the top bar.

ui/src/components/layout/Main.tsx
```diff
@@ -0,0 +1,57 @@
...
+            </UserSpace>
+          </ContainerStyled>
+        </>
+        <DrawerComponent open={open} handleDrawerClose={() => setOpen(false)}/>
+      </BoxStyled>
+    </Container>
+  );
+}
+
+const BoxStyled = styled(Box)`
```
Tbaut, line 49: same comment as for the others (and for `ContainerStyled` bellow), let's use css classes to keep things consistent.

ui/src/components/layout/Main.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <Container maxWidth="lg">
+      <BoxStyled>
+        <Header handleDrawerOpen={() => setOpen(true)}/>
+        <>
+          <ContainerStyled fixed>
+            <UserSpace>
+              <Routes>
```
Tbaut, line 22: Any way we can reuse the `ROUTES` from `src/constants.ts` here? I'm afraid that having the routes duplicated makes it more difficult to maintain.

Tbaut (changes requested):
Super nice, thank you so much for this. I left a couple comments, but we're super close to be golden :)

Tbaut:
BTW Andrew, if you want to take a look and don't want to bother building the branch, you can go to the [deployment page  ](https://github.com/ChainSafe/Multix/deployments/activity_log?environment=multix+%28Preview%29) it's available from the main repo page as well: 
image

And click on the `view deployment`. This will bring you to the deployment of the related commit. I noticed the commit hash don't match though, you should hover it, to see if it's the PR you're looking for.

ui/src/components/Header/index.tsx
Lykhoyda: I can do it for sure, but as a design update are pretty close we probably could wait until the final version

ui/src/components/layout/Main.tsx
Lykhoyda: It's a good idea, I thought to take it to the separate story as it's preferable and more customizable to use createBrowserRouter method so that I will refactor it all together 
https://reactrouter.com/en/main/routers/create-browser-router

asnaith:
Thanks for the tip about deployments @Tbaut 👍

@Lykhoyda I tested this PR out within mobile on both iOS and Android within Nova Wallet. The menu is functioning fine for me but I did notice a couple of other issues whilst viewing on mobile. I've logged them as separate issues (#105 and #106).

ui/src/components/Header/index.tsx
Tbaut: I think it's up to us to define how we want things to look like. While the way you worked here is better to keep things together (if I remove the Typography, the styling will be removed as well), but it's also more code, and makes you jump more when you read it. I've no strong opinion, and I'm not sure we should enforce something, but the rest of the code definitely has a lot of classes, and almost no styled component.

ui/src/components/Header/index.tsx
Tbaut: I think that the design update will really take a while to land, and start, so I think it's a low effort high reward thing to do.

ui/src/components/layout/Main.tsx
Tbaut: ok created an issue for this https://github.com/ChainSafe/Multix/issues/108

ui/src/components/Header/index.tsx
Tbaut: let's keep this as is, and we can go either way, I don't think that's a big deal

Tbaut (approved):

PR #104, opened 2023-04-25
Show identity and info on hover/touch

### LINKED ISSUE #103: Show on-chain identity if available.
Self explanatory I think, particularly useful on kusama, for businesses or DAOs.

Tbaut:
closes #103 
ready to review for a first pass, but still something to do: Use the identity name in the selection dropdown, or anywhere we think it's needed.

@Lykhoyda this is one of the instance where I had to either use either sx, or a styled component, otherwise I don't have access to the DOM node to style it, because it's appended dynamically at the end of the DOM, it's not inside my styled component.

To test, you need to set an identity (e.g you can do this with pjs/apps) https://wiki.polkadot.network/docs/learn-identity

Here's a watch mode on the ChaosDAO where all members have an identity set.
image

alternatively, if you want to see this too, you need to

- point your local node at Kusama (see the .env.example comments in ./squid)
- connect the UI to it using ?network=local
- add a hacky watch address such as "HqfNL8eE5Efvw5Wcj8d3tjeoCS3BPfshMHWG6rmtYUrgyrU" who's part of the ChaosDAO multisig (see the Readme of ./ui to know how to do this)

### DIFF
--- a/ui/src/components/AccountDisplay.tsx
+++ b/ui/src/components/AccountDisplay.tsx
@@ -1,10 +1,13 @@
 import { Box } from "@mui/material";
-import { useMemo } from "react";
+import { useEffect, useMemo, useState } from "react";
 import styled from "styled-components";
 import { useAccountNames } from "../contexts/AccountNamesContext";
 import { AccountBadge } from "../types";
 import { getDisplayAddress } from "../utils/getDisplayAddress";
 import IdenticonBadge from "./IdenticonBadge";
+import { useApi } from "../contexts/ApiContext";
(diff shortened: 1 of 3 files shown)


asnaith (approved):
I noticed that to set an identity on Polkadot you need a reserve of 20 DOT....so I viewed with the `multix.watchedAccount` technique. 

The names are truncated nicely in the mobile view when it gets too small and the hover pop-up is a nice touch too

Tbaut:
I wanted to tweak the account selection to show the name there too, but we'll probably change it soon, so merging as is.

PR #110, opened 2023-05-03
Lykhoyda/migrate workspaces

Lykhoyda:


.github/workflows/lint-squid.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
         key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
     - name: install packages
-      run: npm ci
+      run: yarn install --immutable
```
Tbaut, line 26: This is for subsquid that only officially supports npm, so we need to keep it that way I'd say.
```suggestion
      run: npm ci
```

.github/workflows/lint-squid.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
...
 
     - name: lint
-      run: npm run lint
\ No newline at end of file
+      run: yarn lint
```
Tbaut, line 29: same as above
```suggestion
      run: npm run lint
```

.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
Tbaut, line 1: All these files shouldn't be part of the PR. If you pull main, the gitignore should be fine, and you'll need to update the git cache with `git rm -r --cached .`

Tbaut (commented):
I'm afraid we're going a dangerous route with yarn for subsquid. Let me check with the team if they are very much against supporting yarn before we move on. I messaged them on their telegram group.

.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
Lykhoyda: We are using non-zero installs strategy, which works more reliably than zero installs. According to documentation  plugins  folder could be committed to the repo 
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

Tbaut (approved):
Let's do it, thanks for the effort

Issue #83, opened 2023-04-13
Notification toasts are displayed on top of each other


Tbaut:
**Description**
When several notification toasts are displayed, they are overlaying each other

**Steps To Reproduce**

Add clear steps on how to reproduce the bug

1. Go to Multix at 'https://multix.chainsafe.io/'
2. Initiate a balance transfer
3. When prompted to sign the extrinsic with the extension, cancel it. This will create a toast with the status "error". Toasts error are set to not auto-disappear, this is fine https://github.com/ChainSafe/Multix/blob/867fe5a5f52ea1f3477ce261eb8ef567048c6eaa/ui/src/components/ToastBar.tsx#L29-L30
4. Now if you don't click on the small cross to close it, any other toast will appear on top of it.

Bellow is a screenshot with a long error toast, and a "Tx broadcasted" toast on top of it.

**Expected Behaviour**
the toasts should ideally stack on top of each other.

**Screenshots / Video**
![image](https://user-images.githubusercontent.com/33178835/231721951-359089d8-12bc-433f-b630-579b5b7e0e63.png)

dj-pundir:
Hey @Tbaut  Let me know if I can help you with this. Thanks!!

Tbaut:
Sure, thanks :)

dj-pundir:
Hey @Tbaut  It is not possible to stack the chart with MUI Snackbar. Though they are working on it, once its done we will be able to do that
https://github.com/mui/material-ui/pull/31991

Alternative of this we can use which most of the user used and it seems an alternative by looking at the number of downloads:
https://www.npmjs.com/package/notistack

The original issue of stacking is here
https://github.com/mui/material-ui/issues/1824

Thanks!!

Tbaut:
Notistack is mentioned in the MUI doc yes: https://mui.com/material-ui/react-snackbar/#complementary-projects
We should either use it, or make it work anyhow.

dj-pundir:
Hey @Tbaut I have created customized toast. Here is the sample. Let me know if it works for you. 

[screen-recorder-tue-may-02-2023-17-59-29.webm](https://user-images.githubusercontent.com/61174028/235667194-6c6b9f63-4916-424a-a713-20e049eb8f5c.webm)

It uses MUI Alert where we have 4 options for toast colors. Please check here http://127.0.0.1:3333/create?network=rococo

dj-pundir:
[...]

PR #113, opened 2023-05-05
Add notistack for toast

dj-pundir:
This is the resolution of issue https://github.com/ChainSafe/Multix/issues/83

As discussed i have used `notistack ` where we can show toast on stack up each other.

I have put a folder for newly created stack bar in `~\packages\ui\src\components\SnackBar` (We can place it anywhere else if we want to if this is not the right place)

To change the existing snack bar we only need to do 2 changes

- Put Snackbar in your page

image

- Change existing `useToast()` function to `useSnackStack()` (We can change this in future to sync with existing)

image

- We do not change anything else. The `addToast() ` function will work as usual.

Let me know if it works or you need ore explanation Thanks!!


packages/ui/src/components/SnackBar/SnackToast.tsx
```diff
@@ -0,0 +1,88 @@
...
+    if (toast.duration !== 0) {
+      setTimeout(() => {
+        close();
+      }, toast.duration || 6000);
+    }
+  }, [close, toast.duration]);
+
+  return (
+    <Grow className={className} in={open} timeout={TIMEOUT}>
+      <Alert
```
Tbaut, line 52: Can we please use Snackbar rather than Alerts? Or why do you use Alert? I prefer the look and feel of snackbar, it's more neutral than those loud colors.

packages/ui/src/components/SnackBar/SnackToast.tsx
```diff
@@ -0,0 +1,88 @@
...
+        onClose={handleClose}
+        variant="filled"
+        sx={{
+          minWidth: 300,
+          width: { xs: 1, md: "auto" },
+          mb: 1
+        }}
+        
+      >
+        <div style={{
```
Tbaut, line 64: Please use a className and style it in the `styled(SnackbarToast)` section

packages/ui/src/components/SnackBar/Snackbar.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <MuiSnackbar
+      open={!!firstToast}
+      autoHideDuration={12000}
+      transitionDuration={0}
+      anchorOrigin={{
+        vertical: firstToast?.position?.vertical || "bottom",
+        horizontal: firstToast?.position?.horizontal || "left"
+      }}
+      sx={{
```
Tbaut, line 19: please avoid using `sx` for consistency.

packages/ui/src/pages/Creation/index.tsx
```diff
@@ -36,7 +38,7 @@ const MultisigCreation = ({ className }: Props) => {
   const { selectedSigner, selectedAccount, addressList } = useAccounts()
   const navigate = useNavigate()
   const signCallBack = useSigningCallback({ onSuccess: () => navigate("/?creationInProgress=true") })
-  const { addToast } = useToasts()
+  const { addToast } = useSnackStack()
```
Tbaut, line 41: This should be changed everywhere we use `useToasts`

Tbaut (changes requested):
Thanks a lot for the effort. I made a first pass at the code and left comments:
- once the is working, we should remove all the instances of `useToasts`
- for consistency, please don't use `style=` or `sx`, and only use styled components.
- I'm not sure why you chose to use the Alert component, I think it'd be good to keep the current style

packages/ui/src/components/SnackBar/Snackbar.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const Snackbar: React.FC  = ({ className }: Props) => {
+  const { toastsPack } = useSnackStack();
+  const firstToast = toastsPack[0];
+
+  return (
+    <MuiSnackbar
+      className={className}
+      open={!!firstToast}
+      autoHideDuration={12000}
```
Tbaut, line 17: This doesn't seem to have any effect. Same for the `transitionDuration`. This doesn't sounds needed

packages/ui/src/components/SnackBar/SnackStackProvider.tsx
Tbaut, line 1: This is a context provider, please name it as such, and place it in the `src/contexts`
Feel free to name it `ToastContextProvider` and delete the other one.

packages/ui/src/components/SnackBar/SnackStackProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+  key?: number;
+  title?: string;
+  message?: string;
+  link?: string;
+  duration?: number;
+  type: ToastType;
+  position?: {
+    vertical?: "top" | "bottom";
+    horizontal?: "left" | "right" | "center";
+  };
```
Tbaut, line 15: you can use the type `SnackbarOrigin` right away
```suggestion
  position?: SnackbarOrigin;
```

packages/ui/src/components/SnackBar/SnackStackProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const SnackbarContextProvider  = createContext<SnackStackContextProps>({
+  toastsPack: [],
+  setToastsPack: (toasts) => {},
+  addToast: (toast) => {},
+  removeToast: (key) => {}
+});
+
+const SnackStackProvider: React.FC<React.PropsWithChildren> = ({
```
Tbaut, line 34: Please stay as close as possible to the other conventions the project has: 
```
type ToastContextProps = {
    children: React.ReactNode | React.ReactNode[]
}
```
and then
```
const SnackStackProvider = ({children}: ToastContextProps) => ... 
```

packages/ui/src/components/SnackBar/SnackStackProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+  addToast: (toast: ToastProps) => void;
+  removeToast: (key: ToastProps["key"]) => void;
+};
+
+const SnackbarContextProvider  = createContext<SnackStackContextProps>({
+  toastsPack: [],
+  setToastsPack: (toasts) => {},
+  addToast: (toast) => {},
+  removeToast: (key) => {}
+});
```
Tbaut, line 32: This can be simplified a bit
```suggestion
const SnackbarContextProvider  = createContext<SnackStackContextProps | undefined>(undefined);
```

packages/ui/src/App.tsx
```diff
@@ -10,30 +10,33 @@ import { ThemeProvider as MuiThemeProvider } from '@mui/material/styles'
...
-                    </MultiProxyContextProvider>
-                  </AccountNamesContextProvider>
-                </AccountContextProvider>
-              </ApiContextProvider>
-            </QueryClientProvider>
-          </NetworkContextProvider>
-        </BrowserRouter>
-      </ToastContextProvider>
+      <SnackbarContextProvider>
+        <ToastContextProvider>
```
Tbaut, line 22: please remove the old `ToastContextProvider` altogether, it's no needed any more :)

packages/ui/src/components/SnackBar/Snackbar.tsx
```diff
@@ -0,0 +1,41 @@
...
+        horizontal: firstToast?.position?.horizontal || "left"
+      }}>
+      <Stack flexDirection="column" gap={1}>
+        {toastsPack.map((toast) => (
+          <ToastBar toast={{
+            id: toast.key, 
+            type: toast.type, 
+            title: toast.title ?? '', 
+            link: toast.link }} 
+          key={toast.key} />
```
Tbaut, line 30: There is something not quite right here. Having the `Snackbar` and the `ToasBar` is redundant, and as the other comment suggests, some are not used.

packages/ui/package.json
```diff
@@ -15,6 +15,7 @@
     "graphql": "^16.6.0",
     "graphql-request": "^6.0.0",
     "graphql-ws": "^5.12.1",
+    "notistack": "^3.0.1",
```
Tbaut, line 18: This is not used any more is it?

Tbaut (commented):
Thanks for the work, I left some comments.

Issue #80, opened 2023-04-12
Enhance the manual extrinsics


Tbaut:
While some extrinsics are fairly simple to use in advanced mode, some are not.
e.g `Utility.batch` is basically unusable, or even the `identity.setIdentity` mainly because they expect some encoded data I believe.

We should study the extrinsic tab from github.com/polkadot-js/apps/ repo and try to get to something actually usable.

Tbaut:
I'll go case by case. First one is batch. I may create a dedicated menu entry for it.

Tbaut:
it's outdated and not precise at all. We can reopen specific issues if needed.

PR #114, opened 2023-05-08
Add batch transaction support

Tbaut:
not perfect but it's a start. Already better than what we have, batching is unusable with manual extrinsics.
related to #80 

@asnaith this is far from being perfect, but it allows to do things that I've been asked, namely batching votes, transfers, bounties etc..


asnaith:
I was able to construct a batch transaction and submit/sign successfully. There's more validation and error response handling to come but functionally crafting a batch extrinsic seems good

Tbaut:
Thanks for the review and feedback, I've added:
- a first extrinsic by default
- error handling for batch transactions when they fail, it's very important indeed
- the ability to remove any extrinsic from the batch

asnaith:
I ran through a lot more batch balance transfers and tested out the latest updates here's my feedback

- Latest additions look good 

- It's not possible to reject a `utility.batch` proposal at the moment because only the approval option is shown (for the non-proposing signatory). 

- Minor. The "Remove" button is clickable and perhaps should be greyed out until the first Pallet / Method has been selected as it doesn't function until then. 

Screenshot 2023-05-09 at 3 45 24 PM

- We're not showing the correct text here, in the modal that's displayed after successfully approving a proposal

Screenshot 2023-05-09 at 12 33 05 PM

Potential points for other issues: 

- When extrinsics are stacked in the proposal modal I find that the "remove" and "add extrinsic" options are a little too close together. 

- The display of the params on a proposal in the transaction queue is not great with the additional data from a readability perspective. Perhaps this and the point above are to be left to be fixed in the larger upcoming UI / UX work? 

- Whilst crafting a transaction I had to switch between different apps and another browser to copy addresses. A few times when returning to the browser with multix, I clicked the window to activate it and accidentally clicked out of the modal which lost everything I was setting up for the batch transaction. Perhaps on this modal, we should have a confirmation in this context before closing the dialog? This has always been dismissable like this but could now potentially be a cause for user frustration if they were to lose long batched params that they've been in the middle of crafting.  

- It could be good to have the ability to save and nickname your own custom batch transfers. For example, this might be useful for people making regular payments to the same addresses.

I can create issues for the above if you agree. 

Lastly

- We can't enter decimal point values on batch balance transfers. We have to send 1200000000000 for 1.2 ROC etc. I'm sure this is expected for now but just wanted to mention it.

Tbaut:
Lot's of good things, but indeed not everything can/will be fixed:
> It's not possible to reject a utility.batch proposal at the moment because only the approval option is shown (for the non-proposing signatory).

This is not only for batch, it can't be rejected by anyone else but the proposer. The UX isn't great here, but it's not particular to this PR.

> Minor. The "Remove" button is clickable and perhaps should be greyed out until the first Pallet / Method has been selected as it doesn't function until then.

That's a minor glitch that I'm closing my eyes on. The first method/pallet is added by default, you need at least one otherwise it doesn't make sense as we discussed earlier. When you click on remove at this point, it's actually removed, ...and re-added right away. Checking for the pallet to have changed etc is too much small fiddling code for now (we are using the manual extrinsics component, the code is reused and the potential to create unmaintainable code is high) for me to add it as it'll change in the future. We can create an issue for this, it'll be low prio then and can be tackled when we have more time.

> We're not showing the correct text here, in the modal that's displayed after successfully approving a proposal

Interesting. Do you remember what extrinsic you did to end up with this one? Showing the console could help as well in case you see this, this way I can see the on-chain event.

> The display of the params on a proposal in the transaction queue is not great with the additional data from a readability perspective. Perhaps this and the point above are to be left to be fixed in the larger upcoming UI / UX work?

Yes this is again the same as the manual extrinsic. I need to work on this for sure.

> Perhaps on this modal, we should have a confirmation in this context 

Very good point. I don't think this modal should be an exception, and for the app to feel consistent. Maybe we should change the way modals are displayed.

> It could be good to have the ability to save and nickname your own custom batch transfers. For example, this might be useful for people making regular payments to the same addresses.

I would consider this if this comes from a user with an actual use case. I'll ask.

> We can't enter decimal point values on batch balance transfers.

Yes, this is expected as well. Again it's a manual extrinsic, there's absolutely no help given there. I'm thinking about ways to turn a balance transfer into a batch, from the "send tokens" drop down, where you could add another such transfer. But let's see how useful this will actually be.

asnaith:
> Interesting. Do you remember what extrinsic you did to end up with this one? Showing the console could help as well in case you see this, this way I can see the on-chain event.

@Tbaut This was after doing a batch of balance transfers, was shown after approval (when expecting a success toast message)

Screenshot 2023-05-10 at 4 37 55 PM

I've created other issues for the things you were in agreement with 👍

Tbaut:
Oh thanks I messed up this part. It should be better now, showing any kind of error, and success.

Tbaut:
Yes the finalized notification had been removed for quite a while already. The reason is that the finalization takes some time, and in most cases the in block event is already enough.

PR #117, opened 2023-05-09
Add scripts for docker and remove makefile

Tbaut:


.gitignore
```diff
@@ -119,3 +119,5 @@ build
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+packages/ui/tsconfig.tsbuildinfo
```
Lykhoyda, line 123: It's probably better to ignore this file regardless of the folder
```ssh
*.tsbuildinfo
```

Lykhoyda (changes requested):

.gitignore
Tbaut: agreed

PR #121, opened 2023-05-10
feat: use github actions to deploy on staging and production

rbadillap:
**To deploy on staging:**

Any Pull Request that points to `main` branch, or any commit pushed directly to `main` branch will execute this workflow. The workflow pushes two tags into the docker image: `latest` and the commit SHA. Making an historic list of versions to pick in case of rollbacks.

**To deploy on production:**

Any release created will trigger this workflow. It's important to clarify that the name of the release will be taken to put as a docker tag in the image. So if you plan to create a release, please use a semantic version name in the release, e.g: `v1.0.0`, so I can put that name as the docker tag and push it on production.


rbadillap:
> I have no idea what I'm talking about but from what I can see it looks good. Where is the logic regarding the release BTW, is this somewhere else?

Here's more [information from the official documentation](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository).

And yes, we have other products that manage releases for internal processes.

Screenshot 2023-05-10 at 09 01 19

Other products that worth to be versioned for the public, also uses these type of features, but it doesn't apply in our scenario.
You are free to share your thoughts about another workflow to deploy on STAGE and PRODUCTION if you want.

PR #124, opened 2023-05-10
Show network selector and Kusama

Tbaut:


packages/squid/package.json
```diff
@@ -12,7 +12,7 @@
     "codegen": "npx squid-typeorm-codegen",
     "typegen": "npx squid-substrate-typegen typegen.json",
     "start:indexer": "node -r dotenv/config lib/processor.js",
-    "start:graphql-server": "npx squid-graphql-server --subscriptions --max-response-size 10000 --dumb-cache in-memory --dumb-cache-ttl 1000 --dumb-cache-size 100 --dumb-cache-max-age 1000",
+    "start:graphql-server": "npx squid-graphql-server --subscriptions --dumb-cache in-memory --dumb-cache-ttl 1000 --dumb-cache-size 100 --dumb-cache-max-age 1000",
```
Tbaut, line 15: @rbadillap this is critical to change otherwise we see the "response too big" error

Tbaut: I realized later on that this is not the issue. so you can disregard it.

PR #126, opened 2023-05-10
add prettier config; format all files

### LINKED ISSUE #125: Chore: Add prettier configuration
Align the code formating using prettier tool

Lykhoyda:
closes #125


packages/squid/src/model/generated/account.model.ts
Tbaut, line 1: Same for this one. Ideally, we should take the ignore list from eslint
```
  "ignorePatterns": [
    "db/**/*",
    "src/types/**/*",
    "src/model/generated/**/*"
  ]
```

packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -1,29 +1,43 @@
...
-import { useGetBalance } from "../hooks/useGetBalance";
+import { Box } from '@mui/material';
+import { useEffect, useMemo, useState } from 'react';
+import { styled } from '@mui/material/styles';
+import { useAccountNames } from '../contexts/AccountNamesContext';
+import { AccountBadge } from '../types';
+import { getDisplayAddress } from '../utils';
+import IdenticonBadge from './IdenticonBadge';
+import { useApi } from '../contexts/ApiContext';
+import {
```
Tbaut, line 9: This is the case I mentioned, where I would love to have an exception for import. Making those import on several line doesn't add any value in IMHO

packages/ui/src/components/AccountEditName.tsx
```diff
@@ -1,77 +1,68 @@
...
+  }, [getNamesWithExtension]);
 
-  const onChange = useCallback((event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
-    const value = event.target.value
-    setName(value)
-    onNameChange({ addresses: [address], name: value })
-  }, [address, onNameChange])
+  const onChange = useCallback(
+    (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      const value = event.target.value;
```
Tbaut, line 38: personal opinion, I find `;` useless. Can we add a rule to enforce no semi-colon please?

packages/ui/src/components/AccountEditName.tsx
```diff
@@ -1,77 +1,68 @@
...
-        md={4}
-        alignItems="center"
-      >
-        <AccountDisplay
-          address={address}
-          withName={false}
-        />
+    <Grid className={className} container spacing={2}>
+      <Grid item xs={12} sm={6} md={4} alignItems="center">
+        <AccountDisplay address={address} withName={false} />
```
Tbaut, line 48: Do you know if there's a rule as of why props will be on a different line? I find that starting from more than 1 we should have several lines. but that's my rule of thumb, would love to be able to set it in a config file though.

packages/ui/src/components/AccountSelection.tsx
```diff
@@ -1,104 +1,133 @@
...
-  const [errorMessage, setErrorMessage] = useState("")
-  const ref = useRef<HTMLInputElement>(null)
-  const { accountNames, addName } = useAccountNames()
-  const [name, setName] = useState("")
+  if (!option) return '';
+
+  return typeof option === 'string' ? option : option.address;
+};
+const AccountSelection = ({
+  className,
```
Tbaut, line 46: The function params is a place where I'd like to keep things on one line. This is read very rarely, and bloating the line number doesn't help readability.

Tbaut (commented):
Thanks for putting this in place. As we spoke, let's go for some cases where I disagree with the default formatting on a case by case basis.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "tabWidth": 2,
+  "semi": false,
+  "singleQuote": true,
+  "arrowParens": "avoid"
```
Tbaut, line 5: I'd keep the default always unless you really prefer without.
```suggestion
  "arrowParens": "always"
```

Tbaut (commented):
Looks good for the most part. I guess I can accommodate with the function params. If possible it'd be great if we can remove the trailing comma for objects.

Tbaut (approved):
:rocket: Thanks for handling this :)

PR #128, opened 2023-05-10
Lykhoyda/fix long names multisig

Lykhoyda:
based on #126


packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -99,14 +99,15 @@ const AccountDisplay = ({
...
-      </Box>
+      </BoxStyled>
     </MultisigInfoStyled>
   );
 };
 
 const MultisigInfoStyled = styled('div')`
   display: flex;
   align-items: center;
+  min-width: 0;
```
Tbaut, line 110: This one doesn't seem to be needed. Can you confirm?

PR #140, opened 2023-05-11
feat: use our own github runner to connect to the db to run migrations

rbadillap:
**Custom GitHub Runner**

The intention of this PR is to start using a custom GitHub runner in order to allow the pipelines to connect to our internal database and perform different actions required while deploying our networks, such as, dumping the database and running migrations.

---

We are finally using a separated database per network. Since most of this job was accomplished internally in our infrastructure there's no proof to demonstrate the split, but you already have access to the console and check the logs.

What really matters, is the deploy pipeline, because before deploying, I'm running a db dump and running the migrations back, making sure the schema is fresh.

Let me know if you have questions or needs more clarification.


Dockerfile
```diff
@@ -20,6 +20,7 @@ ADD packages/squid/src ./packages/squid/src
 ADD packages/squid/schema.graphql ./packages/squid
 
 RUN corepack enable
+RUN yarn install
```
Tbaut, line 23: Did you have any issue with that?
We should remove this because this installs all the dependancies, also for the ui, which are not needed.

package.json
```diff
@@ -38,7 +38,7 @@
     "squid:codegen": "yarn workspace multix-squid codegen",
     "squid:typegen": "yarn workspace multix-squid typegen",
     "squid:build": "yarn workspace multix-squid build",
-    "squid:install": "yarn workspaces focus multix-squid",
+    "squid:install": "yarn workspace multix-squid install",
```
Tbaut, line 41: same here, the focus makes sure that only the needed depandancies are installed.

With the current dockerfile, we install twice everything.

yarn.lock
Tbaut, line 1: there has been no change in the dependancies in this PR so there shouldn't be any change in the lock file

Tbaut (commented):
Awesome, left some comments

Dockerfile
Tbaut: I'll merge main into this branch, this should solve any issue you had.

Tbaut:
you are using yarn v1 for the install etc, which doesn't support our workspaces, and that messes up with the lock file. I'll try to make it work with v2

Tbaut (approved):
ok with yarn v2 now I'm good to merge :)

PR #141, opened 2023-05-12
Add a link to the calldata

Tbaut:
The calldata is the most important info about a multisig proposal, it's the call that will be executed. The hash of it is on-chain, and this is what users must be verifying. Multix decodes the call and shows it above (more or less nicely). Providing with a 3rd party link to verify it is a good security addition.

image


asnaith:
This is working great and is a good improvement to user security! 🚀

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -29,6 +31,13 @@ const Transaction = ({
   const isProxy = useMemo(() => isProxyCall(aggregatedData.name), [aggregatedData])
   // FIXME this is duplicated
   const appliedClass = useMemo(() => (isProxy ? 'blue' : 'red'), [isProxy])
+  const { selectedNetworkInfo } = useNetwork()
+
+  const onOpenLink = useCallback(() => {
+    const encodedRpc = encodeURIComponent(selectedNetworkInfo?.rpcUrl || '')
+    const link = `https://cloudflare-ipfs.com/ipns/dotapps.io/?rpc=${encodedRpc}#/extrinsics/decode/${aggregatedData.callData}`
+    window.open(link, '_blank')
+  }, [aggregatedData, selectedNetworkInfo])
```
Lykhoyda, line 40: Why do we need a callback for that? And make a window open instead of using the regular link?
f.e:
```js
function getLink() {
const encodedRpc = encodeURIComponent(selectedNetworkInfo?.rpcUrl || '')
    return `https://cloudflare-ipfs.com/ipns/dotapps.io/?rpc=${encodedRpc}#/extrinsics/decode/${aggregatedData.callData}`
}
```

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -57,9 +66,21 @@ const Transaction = ({
         aggregatedData={aggregatedData}
         children={
           (isProposer || possibleSigners.length > 0) && (
-            <div className="buttonWrapper">
+            <TransactionFooterStyled>
+              {!!aggregatedData.callData && (
+                <Linkstyled
+                  className="linkIcon"
+                  onClick={onOpenLink}
```
Lykhoyda, line 73: I think it's better to use Link component from Mui library instead of box for the links:
```js
                <Linkstyled
                  className="linkIcon"
                  href={getLink()}
                  target="_blank"
                >
                  See callData
                  <LaunchIcon
                    className="icon"
                    fontSize="small"
                  />
                </Linkstyled>
```

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -75,21 +96,37 @@ const Transaction = ({
   )
 }
 
+const Linkstyled = styled(Box)(
```
Lykhoyda, line 99: Link component

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -29,6 +31,13 @@ const Transaction = ({
   const isProxy = useMemo(() => isProxyCall(aggregatedData.name), [aggregatedData])
   // FIXME this is duplicated
   const appliedClass = useMemo(() => (isProxy ? 'blue' : 'red'), [isProxy])
```
Lykhoyda, line 33: I think it's unnecessary in this case to useMemo as it's a simple operation. It will only slow down the render in this case and create confusion. useMemo strives when we sort or filter big arrays of data or some heavy computation. But IMO here is not a case as well for isProxyCall

Lykhoyda (dismissed):
The functionality works as expected. Some little changes could improve the code readability

packages/ui/src/components/Transactions/Transaction.tsx
Tbaut: sounds good

packages/ui/src/components/Transactions/Transaction.tsx
Tbaut: No particular reason, just used to it. I've set the link in a variable instead

packages/ui/src/components/Transactions/Transaction.tsx
Tbaut: I literally never use `const` in the code, and preferred `useMemo`. This is certainly overkill in this case, but as a rule of thumb, I don't want to be discussing performance for a certain variable, and not for another. So I chose to have useMemo all the time and I don't think this will hurt the performances. This is at the cost of readability that's true, however I do not want to open this box of worms.

Lykhoyda: I didn't get the reasoning behind using useMemo based on the const. If you agree that is overkill and according to the documentation, it's not a case for useMemo, then which problem we are trying to solve by wrapping it? My point is that simple operations like checking the boolean value or length shouldn't be used with the hook, because instead of checking the easy operation we wrap it with useMemo method which makes more checks than the original code. Plus when I see something wrapped in the hook it makes me think that is a heavy computation operation, but not in that case. The React will re-calculate the value anyway if the props will be changed.

Tbaut: My point is what the doc says:
> There is no benefit to wrapping a calculation in useMemo in other cases. There is no significant harm to doing that either, so some teams choose to not think about individual cases, and memoize as much as possible. The downside of this approach is that code becomes less readable. 

I'm in the camp, because of habit. In the whole repo, you'll probably never see in a react component:
```
const someVar = [some logic]
```
I've always wrapped them, as a habit
```
const someVar = useMemo(() => [some logic] ,[])
```
Because "There is no significant harm to doing that either", I don't want to be discussing the `useMemo` for variables. I wrap them all without distinction. Just like Prettier isn't perfect, it's opinionated so as not to spark discussions for styling.

Lykhoyda: Yes, indeed they wrote that there is no harm in one sentence, but also mentioned the following:
```text
The downside of this approach is that code becomes less readable. 
Also, not all memoization is effective: a single value that’s “always new”
is enough to break memoization for an entire component.
```
So they agree that the code will still work both ways, but it's not meaning that it's becoming more efficient, and in some cases, it does more harm than adding value. If you are not against reading one article that explains the cases for the usage of the hooks and downsides. If you are not convinced after that then fine we are keeping it
https://kentcdodds.com/blog/usememo-and-usecallback

Tbaut: Thanks for the link. I've re-read it although I did in the past. The cost to me is negligible, while the pros are very clear: 
-> if a variable breaks the memoization then the worse that's happening is that it's useless. In other cases, it actually helps, and there's no discussion about it or forgetting when refactoring that you should add/remove the memoization. The direct benefit is that there's no infinite loop to check, which is something very easy to overlook in a code review. The pros really outweigh the cons for me.

asnaith (approved):

Tbaut:
merging, it's been lagging here for too long.

PR #146, opened 2023-05-15
format long balance

### LINKED ISSUE #67: Show more decimal places on the signing proposal modal
We currently only show the first 4 numbers after the decimal point which means proposals of 0.00009 will be displayed as "0.0000" (although the correct amount is still sent if it's approved). 

It would be good to expand this to show the lowest supported amount. 

Screenshot 2023-03-28 at 2 52 15 PM

Lykhoyda:
closes #67


asnaith:
This is working well. Great improvement to the home page balances and signing proposals.

asnaith (approved):

Tbaut (approved):
The code looks good to me, thanks.

Issue #92, opened 2023-04-14
Show "Copied" info when an identicon is clicked


Tbaut:
Input from users:

I don't understand why the cursor of the mouse has a "+" when it's hovering an identicon.
This is a behavior from the @polkadot/react-identicon repo. If clicked, the address of the account will be copied to the clipboard. 
Users need to understand this.

Proposal: show a "copied" tool-tip above or below the identicon when clicked. It should be visible for ~5 sec
Alternatively we could use a toast with the toast context `useToasts`

tarunsamanta2k20:
I'd like to work on this issue

Tbaut:
Thanks for proposing, feel free to.

tarunsamanta2k20:
@Tbaut For which indenticon I have to make changes?

Tbaut:
all of them!

the best way is to:
- make a dedicated Identicon component that shows the copy info and pass all other props to the current Identicon implementation
- replace all the occurrences of the `Identicon` component

tarunsamanta2k20:
![image](https://user-images.githubusercontent.com/55488549/233816212-39bd523e-11ba-47f2-85aa-aed53051b94c.png)


which one is tooltip here?

tarunsamanta2k20:
@Tbaut tooltip or useToast?

Tbaut:
I think tooltip would be better. Thanks.

tarunsamanta2k20:
I  have implemented the feature I don't know why it is not working. Shall I push it? Could you please check it once?

Tbaut:
You can push, but please provide some info around what you did and what doesn't work.

Tbaut:
Hey, absolutely

PR #149, opened 2023-05-16
Add tooltip for identicon

dj-pundir:
This is the resolution of issue https://github.com/ChainSafe/Multix/issues/92

Created new class `MultixIdenticon` for tooltip purpose.

I have done following more change:
- Remove theme and size properties : Because most of the places they are using the same values so i made them optional parameters in the control and uses fixed values `ICON_THEME, ICON_SIZE` if not pass.
- Please check the `DEFAULT_AUTO_HIDE_DURATION` time as 2 secs look ok for me as you explained in the requirement it should be around 5 secs. Please change if you want it otherwise.
- Placement can be `bottom` also which can be changed by `DEFAULT_PLACEMENT`

image


packages/ui/src/components/AccountSelect/index.tsx
```diff
@@ -36,10 +36,9 @@ export const AccountSelect: React.FC<Props> = ({ anchorEl, onClose, ...props })
           <CardHeader
             avatar={
               <IconButton sx={{ p: 0 }}>
-                <Identicon
+                <MultixIdenticon
                   value={account.address}
                   theme="substrate"
```
Tbaut, line 41: We should remove the theme here too I believe
```suggestion
```

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface Props {
+  value?: string
+  theme?: IconTheme
+  size?: number
+  className?: string
+}
+const MultixIdenticon = ({ value, theme = ICON_THEME, size = ICON_SIZE, className }: Props) => {
+  const [open, setOpen] = React.useState(false)
+  const handleTooltipClose = () => setOpen(false)
```
Tbaut, line 20: Please use `useCallback`here and below
```suggestion
  const handleTooltipClose = useCallback(() => setOpen(false), [])
```

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
...
+const DEFAULT_AUTO_HIDE_DURATION = 2000
+
+interface Props {
+  value?: string
+  theme?: IconTheme
+  size?: number
+  className?: string
+}
+const MultixIdenticon = ({ value, theme = ICON_THEME, size = ICON_SIZE, className }: Props) => {
+  const [open, setOpen] = React.useState(false)
```
Tbaut, line 19: Let's import the `useState` from `react` directly.
```suggestion
  const [open, setOpen] = useState(false)
```

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
```
Tbaut, line 1:
```suggestion
import { useState, useCallback } from 'react'
```

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface Props {
+  value?: string
+  theme?: IconTheme
+  size?: number
+  className?: string
+}
+const MultixIdenticon = ({ value, theme = ICON_THEME, size = ICON_SIZE, className }: Props) => {
+  const [open, setOpen] = React.useState(false)
+  const handleTooltipClose = () => setOpen(false)
+  const handleTooltipOpen = () => setOpen(true)
```
Tbaut, line 21: using useCallback here too
```suggestion
  const handleTooltipOpen = useCallback(() => setOpen(true), [])
```

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <Tooltip
+      PopperProps={{ disablePortal: true }}
+      onClose={handleTooltipClose}
+      open={open}
+      title={DEFAULT_TITLE}
+      placement={DEFAULT_PLACEMENT}
+      leaveDelay={DEFAULT_AUTO_HIDE_DURATION}
+      arrow
```
Tbaut, line 31: I'd suggest to remove the arrow.
```suggestion
```

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
...
+    >
+      <IconButton onClick={handleTooltipOpen}>
```
Tbaut, line 33: Is this needed or can we have the `onClick` handler on the Identicon directly?

packages/ui/src/components/MultixIdenticon.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import Tooltip from '@mui/material/Tooltip'
+import { ICON_SIZE, ICON_THEME } from '../constants'
+import Identicon from '@polkadot/react-identicon'
+import { IconButton } from '@mui/material'
+import { IconTheme } from '@polkadot/react-identicon/types'
+
+const DEFAULT_PLACEMENT = 'top'
+const DEFAULT_TITLE = 'Copied'
```
Tbaut, line 9: Maybe it would be clearer to our users what is actually copied
```suggestion
const DEFAULT_TITLE = 'Address copied!'
```

Tbaut (changes requested):
Thank you so much for your contribution. This looks very good. I left some small comments, but nothing critical.

packages/ui/src/components/MultixIdenticon.tsx
dj-pundir: Hey @Tbaut,  Is it always ADDRESS what we copy throughout the app?

dj-pundir: Hey @Tbaut  Also Is there any way i can click on CREATE button here. I am sure we are using rococo testing environment here to send ROC etc.

image

Tbaut: The error says you don't have enough ROC on the Signing account. Get some ROC using the faucet: https://matrix.to/#/#rococo-faucet:matrix.org

> Hey @Tbaut, Is it always ADDRESS what we copy throughout the app?

The copy action is from the `Identicon` component that we take from a library, so we don't handle the copy per se, and yet it'll always be the address copied in this case

dj-pundir: > The error says you don't have enough ROC on the Signing account. Get some ROC using the faucet: https://matrix.to/#/#rococo-faucet:matrix.org


@Tbaut  it is showing choose app to continue. Can i choose any one of them?

image

asnaith: I can answer this one @dj-pundir. Yep, you should be able to use any of them. Element seems to be the most popular.

packages/ui/src/components/MultixIdenticon.tsx
Tbaut: I'd appreciate an answer to this question

Tbaut:
Please make sure it's passing the lint and building without warning.

Tbaut (changes requested):
Please sure to run `yarn lint --fix` and resolve any issue.

packages/ui/src/components/MultixIdenticon.tsx
dj-pundir: Hey @Tbaut  sorry i guess i missed this.
So if we are using tooltip on `Icon` then we have to create a button of that `Icon` and need a wrapper around (using `<IconButton>` ) to use `Tooltip` but for Button there is not need of any wrapper like below: (Ref: https://mui.com/material-ui/react-tooltip/)

image

About using onClick() directly into Identicon we can do that. Though the method will change to onCopy() but works fine.

image

Let me know if we should change that. Thanks!!

dj-pundir:
> Please sure to run `yarn lint --fix` and resolve any issue.

Actually when i ran yarn lint --fix its showing so many problems at my end so i fixed the problems only for my files which i had changed. Not sure why this is happening. Thanks!!

image

tarunsamanta2k20:
@Tbaut is this issue resolved?

packages/ui/src/components/MultixIdenticon.tsx
Tbaut: Yes, let's use `onCopy` then. No need to wrap it.

Tbaut:
Please run `yarn formatAll` at the root of the package, and `yarn lint --fix` too. Thanks

packages/ui/src/components/MultixIdenticon.tsx
dj-pundir: Sorry but we have to use `<IconButton>` wrapper even if we use onCopy() handler. Because `Identicon` is an icon type not a button and `<Tooltip>` understands button only. I have committed the changes please check. Thanks!!

Tbaut: The current state of the PR is not acceptable, it breaks the UI
image

Tbaut: You can change this IconButton by a div. Also please make sure the tooltip is on 1 line.
image

dj-pundir: Please check now its using same style as before. Thanks!!

[screen-recorder-mon-may-22-2023-16-34-34.webm](https://github.com/ChainSafe/Multix/assets/61174028/80aa466c-29a5-4083-976c-0f8bc88ef5bc)

Tbaut (approved):
Awesome, thanks a lot

Issue #38, opened 2023-03-09
Show who still needs to sign


Tbaut:
When a multisig porposal is available, the info available on-chain are:
- who is the proposer
- what is the threshold
- what's the hash of the proposal
- who has approved

Note that there's no way to reject/cancel a proposal unless you are the proposer. There's an anti-spam measure in the sense that some funds are reserved for any proposal.

The UI currently has the following flow:
- it gets a notification that a new proposal is available for the current multisig thanks to our indexer subscription. 
- `usePendingTx` gets the on-chain pending transactions for our multisig
-  `getAgregatedDataPromise` goes and fetch the on-chain data for the call (based on the block height that we got from the previous step). The info it gets is mostly for users to be able to review the multisig proposals.
- We display the proposal information (what call)
 
Now, we do have the info about the possible signers (who are the signers of the multisig), but we don't display yet who has already signed, and who hasn't. Or how far we are in the signing process.

We need to show this in a smart way, that is compatible with mobile (info on mouse hover isn't great). Some idea was using a slide bar at the bottom of the proposal (some div with a backrgound or just the borders, something like:

![image](https://user-images.githubusercontent.com/33178835/231580738-941f2633-fedc-4358-9a46-ba8ceace4755.png)

PR #153, opened 2023-05-17
add progress bar for pending transactions

Lykhoyda:
closes https://github.com/ChainSafe/Multix/issues/38


Tbaut (commented):
Thanks. This is quite a good addition but the main point of the original issue is to know still need to sign. This is really something that we should show somehow.

I'm thinking about a new design that sounds more clickable, which would be a circular progress, and on click you'd show in a tooltip "Signature required:" and the list of those who didn't sign.

Lykhoyda:
> 

I thought we agreed that the detailed information about who needs to sign would be on the detailed page of the transaction as it will be too much information to display (the new design will solve the issue). The user can quickly see the pending status of the transaction and know how much is left. Because what you can do with the information that account 3 didn't sign the transaction if you are not an owner of the account? (There is no user flow in our app on what to do with this information). If it's important to show as a developer experience it's better than to show some list with pending signatories and not to hide the info in the tooltips. I find it extremely unreadable for such a huge string to be in tooltips.

Lykhoyda:
> > 
> 
> @Tbaut I thought we agreed that the detailed information about who needs to sign would be on the detailed page of the transaction as it will be too much information to display (the new design will solve the issue). The user can quickly see the pending status of the transaction and know how much is left. Because what you can do with the information that account 3 didn't sign the transaction if you are not an owner of the account? (There is no user flow in our app on what to do with this information). If it's important to show as a developer experience it's better than to show some list with pending signatories and not to hide the info in the tooltips. I find it extremely unreadable for such a huge string to be in tooltips.

BTW what about following option?
Screenshot 2023-05-22 at 16 22 27

Lykhoyda:
> You can ping ppl offline and tell them to sign.

Isn't it the problem we are trying to solve to get rid of off-chain communication?

Tbaut:
Knowing who is lagging to sign and letting them know **again** is different than having to pass them a string of hex. BTW This is not an issue I came up with. This is real feedback, from real users.

asnaith:
Hey @Lykhoyda please add me as a reviewer when the final changes are in if you want me to test out the functionality :)

asnaith:
@Lykhoyda I want to spend some more time looking at the pending status but running out of time today so will pick it up again on Wednesday. I did however find some instances where the layout is breaking: 

Screenshot 2023-05-23 at 2 19 23 PM

Screenshot 2023-05-23 at 2 23 20 PM

packages/ui/src/components/Transactions/TransactionProgress.tsx
```diff
@@ -0,0 +1,128 @@
...
+import HourglassBottomIcon from '@mui/icons-material/HourglassBottom'
+import AccountDisplay from '../AccountDisplay'
+import CheckIcon from '@mui/icons-material/Check'
+
+interface TransactionProgressProps {
+  value: number
+  threshold: number
+  possibleSigners: string[]
+  approvals: string[]
+}
```
Tbaut, line 24:
```suggestion
}

```

packages/ui/src/components/Transactions/TransactionProgress.tsx
```diff
@@ -0,0 +1,128 @@
...
+const ListItemToSign = ({ approvals, signer }: { approvals: string[]; signer: string }) => {
+  return (
+    <ListItem disablePadding>
+      <ListItemButton>
+        <ListItemIcon>
+          {approvals.includes(signer) ? <CheckIconStyled /> : <HourglassBottomIcon />}
```
Tbaut, line 30: At first I thought, let's show only the ones who didn't sign, but I guess it's the same info more or less. Let's go with this. I must remember to ask users what they prefer and if that works.

yarn.lock
Tbaut, line 1: There was no change in the deps, can we keep the current yarn.lock?

Tbaut (commented):
There's something weird with the modified yarn.lock. When run locally, the UI doesn't display and there's a console.error 
image

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -75,19 +83,23 @@ const Transaction = ({
   )
 }
 
+const TransactionCallInfoBoxStyled = styled(Box)`
+  width: 100%;
+`
+
```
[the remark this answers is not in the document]
Lykhoyda: please double check as I can see only one line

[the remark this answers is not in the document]
Tbaut: aaah maybe I need some glassed. I didn't see the `

Tbaut:
In case that helps I can see what's wrong with the UI breaking. It's linked to the CallInfo that's overflowing.
image

Lykhoyda:
> In case that helps I can see what's wrong with the UI breaking. It's linked to the CallInfo that's overflowing.
> 
> image
> 
> 

How can I reproduce such long parameters?

Tbaut:
This is a batch call, you can do one with the following. The dest is any address, and the value is in plank.
image

asnaith:
> There's something weird with the modified yarn.lock. When run locally, the UI doesn't display and there's a console.error

I'm also seeing this locally. After manually deleting the yarn.lock file and running yarn again to generate a new file, Multix doesn't render properly and in the console, I see that same error

asnaith:
Hey @Lykhoyda, @Tbaut's batch tx example all is all you need to reproduce the layout issue - as soon as you have a proposal with those longer params you will see it. It doesn't matter what viewport (it happens at 720p - 4K resolutions) and not browser specific (happens both on Firefox and Chrome).

Lykhoyda:
> 

It looks like there is a new issue with vite related to the packages caching https://github.com/vitejs/vite/issues/12423
It also sometimes doesn't render the page for me.

Lykhoyda:
@asnaith @Tbaut please check again the UI. It should work now :) 
Btw if you have an issue with starting the UI it's probably because of the vite cache. Try to remove folder /packages/ui/src/node_modules/.vite  The issue is open in vite github https://github.com/vitejs/vite/issues/12423

packages/ui/src/components/Header/index.tsx
```diff
@@ -12,22 +12,18 @@ import { isEmptyArray } from '../../utils'
...
   const isAccountConnected = useMemo(() => !isEmptyArray(ownAccountList), [ownAccountList])
   const isAtLeastOneMultiProxy = useMemo(() => !isEmptyArray(multiProxyList), [multiProxyList])
   const { isAllowedToConnectToExtension, allowConnectionToExtension } = useAccounts()
 
   return (
-    <MuiAppBar
-      position="fixed"
-      className={className}
-    >
+    <MuiAppBar position="sticky">
```
Tbaut, line 26: this changed the UI quite a bit, it's off now
image

packages/ui/src/components/Header/index.tsx
```diff
@@ -118,4 +112,39 @@ const IconButtonStyled = styled(IconButton)(
 `
 )
 
-export default Header
+const LinkStyled = styled(Link)(
+  ({ theme }) => `
+  color: white;
```
Tbaut, line 117: can I ask where this all comes from? The white should probably come form the theme.pallet.primary.white or something

Tbaut (commented):
Any clue/easy way to prevent the line from jumping?

https://github.com/ChainSafe/Multix/assets/33178835/12181173-5f83-4c0f-b941-ca3f534a063d

asnaith:
> @asnaith @Tbaut please check again the UI. It should work now :) Btw if you have an issue with starting the UI it's probably because of the vite cache. Try to remove folder /packages/ui/src/node_modules/.vite The issue is open in vite github [vitejs/vite#12423](https://github.com/vitejs/vite/issues/12423)

Oh thanks for the tip. Yep, I'm not seeing the ui rendering problem anymore and I do see that the layout is no longer breaking on the calls with long params 👍

Tbaut:
Another styling issue I realized when playing with it. The names etc appear to be clickable (there's a hover styling, the cursor is pointer) but it effectively does nothing when clicked. Can we remove this styling so as not to confuse users please?

image

packages/ui/src/components/Header/index.tsx
Lykhoyda: I styled the Link from react router similar to the Button component from mui library as we are not using the class anymore.

Tbaut (changes requested):

packages/ui/src/components/Transactions/TransactionList.tsx
```diff
@@ -181,7 +181,8 @@ const TransactionList = ({ className }: Props) => {
               aggregatedData={agg}
               isProposer={isProposer}
               onSuccess={refresh}
-              possibleSigners={possibleSigners}
+              possibleSigners={multisigSignatories}
```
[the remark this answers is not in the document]
Tbaut: Yes, my previous test now shows the right amount of signatories. I'll test some more just to be sure and merge
image

PR #161, opened 2023-05-22
Add watch account functionnality

### LINKED ISSUE #145: Watch account
While we have a hacky way to watch an account using local storage, we should allow to do it using the UI
- Add a Setting menu
- Add a Account Management section
- Show the watched account **signers** and allow to add/remove one
- Only support adding signers for the start, we should have another issue to support adding multisigs
- Make sure we have some metadata associated to the account so that we don't treat it as if we were able to sign with it

Things to test:
- make sure we see the multisigs associated with the watched account
- when an account is removed the query should be fired again without the account
- I shouldn't be able to sign/review a proposal if I only have a watched account

Tbaut:
closes #145 
- find it in Settings>Watch account or on the home page if no account was created.
- supports watching a pure, a multisig or any signatory

Here are signatories accounts to watch on 
- Kusama: `HqfNL8eE5Efvw5Wcj8d3tjeoCS3BPfshMHWG6rmtYUrgyrU`
- Rococo: `5Fer9CC593nS2JTsS4SFC3iWT7M2cwcVWqixjAQ9uPDKg8Jg`

- Pure proxy on Kusama "ChaosDAO Referendum Commission"
`DCZyhphXsRLcW84G9WmWEXtAA8DKGtVGSFZLJYty8Ajjyfa`
- Multisig on Kusama 
`CbsxftwHVqZHEdcv1377x5SGoDF9oC5Knvq3BbDsnTPk99L`

image


asnaith:
Nice to have a UI for this 🙌! 

I've found a few minor issues / suggestions : 

- The nickname for the watched account does not persist between networks 

https://github.com/ChainSafe/Multix/assets/3982665/7972f977-ac2d-40f3-890f-9d49cfb74818

- After adding a signatory to the watched list Multix will also display all the multisigs and pures that the signatory is behind in the dropdown selector. I understand why this is happening but could be confusing to the user? 

https://github.com/ChainSafe/Multix/assets/3982665/354c5f25-d52a-4e4f-8764-15fb8db8ef7f

- If the account you are watching already has an identity set then this will supersede the nickname and also show up twice in the dropdown under its identity and the nickname. Perhaps the nickname should take precedence as the user is specifically naming it?

https://github.com/ChainSafe/Multix/assets/3982665/4a01dc66-fc74-41e5-adae-d249de2c2afb

- When there are accounts that have been added "Watched" makes more sense then "Watch" in this context 

Screenshot 2023-05-23 at 1 30 09 PM

Tbaut:
Thanks for the review
- regarding all the multisigs and pure behind the signatory, this is by design, it's actually a feature :) After all, a multisig could be the controller of many pure proxies etc. There's no way to know what multisig/pure proxy a user want to see.
I think the confusion could be lowered if we could show clearly that the reason an account is in the dropdown is because of a watched account? The only way I would find easy to understand would be with :eyes: emoji next to the identicon or something :)

All the other comments are totally valid, just not related to this PR. Regarding the nickname, I think the identity should take precedence. The multiproxy selection issue is something I noticed when I added the feature, but left undone (searching by identity is not really possible unfortunately)

Tbaut:
I'd appreciate if you can give this one another go @asnaith  with a focus on
- extension connection (you can play with the localStorage value `multix.canConnectToExtension` which should be se to false when a user visits Multix for the first time and never click on "Connect". We will never fire a connection request to the extension before this is set to `true`, meaning before explicit request from the user.
- extension connected but not sharing any account
- extension sharing an account that has no multisig
^ all these scenarios should still allow users to watch and see other accounts/multisigs

I did test thoroughly this all, but better double check. Also say a user doesn't connect the extension but watches an account with mutlisgs, they should still be allowed to connect their extension (with the new Connect button at the top-right). This menu should be consistent in mobile as well (the code for this isn't shared, so worth checking). 
Finally, we should never propose users to create a new multisig if they don't share at least an account with us.
Again something I tested thoroughly, but I may have missed something.

asnaith:
Hey @Tbaut, I checked out the latest changes as requested. The scenarios below are all checking out successfully for me too: 

- User should never be prompted to connect extension unless they have clicked the "connect wallet" link or "Connect" button (to set local storage multix.canConnectToExtension to true). 

- User should be notified of "No account found" when choosing to connect their wallet to multix but not sharing an account within the extension

- User can watch an account without connecting their wallet extension to Multix

- User that was previously only watching an account can connect their own account and continue to use multix

- User is notified "No multisig found for your accounts" when connecting wallet address to multix with no pre-existing multisig

I checked out the mobile view and the connect button on the slide-out menu is behaving consistently too. 

There's only one thing that could potentially confuse new users: If you are wanting to watch an account but have defaulted to the wrong network without realizing it. In the video below I was defaulting to Rococo, I added a signatory from an account on Kusama to the watch list but until I was on the correct network nothing was displayed (expected but potentially confusing). Could it be worth asking the user to choose the network as they are adding the watched account and then automatically update the connection rather than rely on them to choose the network in another component first / after?

https://github.com/ChainSafe/Multix/assets/3982665/e166214b-8cde-4b84-9bc3-2180c35140b0

Tbaut:
ok, while I don't think we should actually let users select yet another network (it's pretty clear what network we're on). I'll make sure that we store and show the address in the right format depending on the network. I'll do this in a follow-up PR and solve the nickname issue at the same time.

Tbaut:
Assuming this is approved. merging.

PR #172, opened 2023-05-24
Fix docker:start:all script

Tbaut:
self explantory


asnaith:
Thank you for the fix!

asnaith (approved):

PR #181, opened 2023-05-30
add Jost font

### LINKED ISSUE #166: Adapt the font to the new design
The titles have not been super consistent and I've not used Typography at all. This issue could also tackle this if wanted
`font-family: 'Jost';`

image

Lykhoyda:
Closes #166


package.json
```diff
@@ -47,5 +47,8 @@
     "eslint-config-prettier": "^8.8.0",
     "eslint-plugin-prettier": "^4.2.1",
     "prettier": "^2.8.8"
+  },
+  "dependencies": {
```
Tbaut, line 51: I believe this doesn't need to be installed for all the packages since it's only present in the `ui`. I'd suggest to add it to the `packages/ui` deps rather than the global one.

Tbaut (approved):
Awesome, thanks for the cleanup as well

PR #188, opened 2023-06-02
Show and group by date of proposal

### LINKED ISSUE #82: Display multisig proposals sorted by time
Right now, I the proposals are displayed randomly (it seems).
We have the information for each proposal regarding when it was actually proposed.
It would be nice to
- sort the proposals by date
- show the date

While we may not have the date from the indexer (the `MultisigCall` entity has it but the source of truth should be the blockchain). We have the ability to query this information for the block in which the call was made, there should be a `timestamp` param.

Tbaut:
closes #82 

image


asnaith:
Really nice improvement that's working well 🚀

asnaith (approved):

Lykhoyda (approved):

PR #197, opened 2023-06-07
Fix nits

Tbaut:
- remove warning that mapped elements need a key
- we were showing "Multisigs" even if we only had 1.


asnaith:
Good point regarding "Multisig" / "Multisigs"! I've tested it out, working well.

PR #201, opened 2023-06-08
update buttons and links

### LINKED ISSUE #200: Update all buttons design in the project
Scenario: All the buttons and links components should be updated with the new UI component from #189 
Given: All pages of the project
When: User interacts with Button or Link
Then: The design should correspond to the new design from Figma https://www.figma.com/file/4dQC9kxf5uxCrAjpkzAK0D/Multix?type=design&node-id=229-12512&t=NpmL9FmHrgq6g8Yi-0

Lykhoyda:
Closes #200

### DIFF
--- a/packages/ui/src/components/CallInfo.tsx
+++ b/packages/ui/src/components/CallInfo.tsx
@@ -155,10 +155,6 @@ const CallNameStyled = styled('h4')`
 const Linkstyled = styled(Link)(
   ({ theme }) => `
   display: flex;
-
-  .icon {
-    margin-left: .5rem;
-  }
 `
 )
 
--- a/packages/ui/src/components/CallInfo.tsx
+++ b/packages/ui/src/components/CallInfo.tsx
@@ -9,8 +9,8 @@ import { formatBnBalance } from '../utils/formatBnBalance'
 import MultisigCompactDisplay from './MultisigCompactDisplay'
 import { Launch as LaunchIcon } from '@mui/icons-material'
 import { useNetwork } from '../contexts/NetworkContext'
-import { Link } from '@mui/material'
 import NoCallInfo from './NoCallInfo'
+import { Link } from './library'
 
 interface Props {
   aggregatedData: AggregatedData
--- a/packages/ui/src/components/CallInfo.tsx
+++ b/packages/ui/src/components/CallInfo.tsx
@@ -155,6 +155,7 @@ const CallNameStyled = styled('h4')`
 const Linkstyled = styled(Link)(
   ({ theme }) => `
   display: flex;
+  padding-left: 0.5rem;
 `
 )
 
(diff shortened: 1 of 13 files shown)


Tbaut (commented):
Awesome, thanks for this. Here are some comments
- Can we have the font-size of buttons be 16px, this corresponds to the medium ones.
- let's use 2 buttons here, and fix the connect at the top-right. to test, you can use the deployment link, since it's a new domain, you won't be connected. You can also change the localStorage variable called `multix.canConnectToExtension` to `false`.
image

- the link icon isn't aligned properly
image

- it's not on the design yet, but discussed this with @sweetpea22 and I think it makes sense to have the back be secondary. As a rule of thumb I guess having 1 primary button per page is good.
image

- in Modals too, if we have 2 buttons, then the "back" should be secondary. The next/save/create should be primary.

Lykhoyda:
> Awesome, thanks for this. Here are some comments
> 
> * Can we have the font-size of buttons be 16px, this corresponds to the medium ones.
> * let's use 2 buttons here, and fix the connect at the top-right. to test, you can use the deployment link, since it's a new domain, you won't be connected. You can also change the localStorage variable called `multix.canConnectToExtension` to `false`.
>   image
> * the link icon isn't aligned properly
>   image
> * it's not on the design yet, but discussed this with @sweetpea22 and I think it makes sense to have the back be secondary. As a rule of thumb I guess having 1 primary button per page is good.
>   image
> * in Modals too, if we have 2 buttons, then the "back" should be secondary. The next/save/create should be primary.

@Tbaut thank you for the review
1. button and connect fixed
2. Icon aligned
3. 'Back' changed to secondary

Tbaut:
Thanks, there are a couple things still:
- I think you forgot that, can we have the font-size of buttons be 16px for buttons, this corresponds to the medium ones.
- When the modals have only 1 button, I guess it should be primary
- I saw in the "change multisig" modal has a back/save where both are still secondary
- The link icon is super far from the title, maybe we should have a padding-left of only .5rem
image

Lykhoyda:
> Thanks, there are a couple things still:
> 
> * I think you forgot that, can we have the font-size of buttons be 16px for buttons, this corresponds to the medium ones.
> * When the modals have only 1 button, I guess it should be primary
> * I saw in the "change multisig" modal has a back/save where both are still secondary
> * The link icon is super far from the title, maybe we should have a padding-left of only .5rem
>   image

@Tbaut updated
regarding the "I saw in the "change multisig" modal has a back/save where both are still secondary", it could be a cache as it works on my side 
Screenshot 2023-06-09 at 12 04 46

Tbaut (approved):
Thanks a lot. 
I pushed some more nit fixing that were quicker to do than to comment about
- The approve button was still not primary (the reject was, but they never appear together)
- The buttons for batch extrinsics were too close together, added 1rem space. 
- The tx list was super close to the multisig list, not sure where this happened, certainly not this PR, anyway I added 1.5rem when on larger screens
image

PR #208, opened 2023-06-14
add phala and khala networks

mpetrunic:


Tbaut (approved):
Let's go, thank you so much for that :heart:

PR #209, opened 2023-06-14
Don't show 'send' if watched pure

Tbaut:
self explanatory, caused a crash if clicking on "send" on the watched account.

try it out, watch DCZyhphXsRLcW84G9WmWEXtAA8DKGtVGSFZLJYty8Ajjyfa


asnaith (approved):
This looks good to go, working as intended 👍

PR #212, opened 2023-06-15
fix inputs misalignment

### LINKED ISSUE #211: Inputs become vertically misaligned when error message is displayed
**Description**
When an error message appears it causes the input to be raised and become vertically misaligned with other inputs in the layout

**Steps To Reproduce**

Invoke an error message eg by trying to add an invalid address to a new multisig

**Expected Behaviour**
The error message should appear under the input without causing it to be vertically misaligned. 

**Screenshots / Video**

Example: 

Screenshot 2023-06-14 at 10 51 15 AM

**Test Host Specifics: **

- Not specific to an OS / Browser

Lykhoyda:
Closes #211


Tbaut (approved):
awesome, thanks a lot!

PR #216, opened 2023-06-16
Allow forming extrinsics from a pjs callData

Tbaut:
Adds a transaction possibility allowing users to paste a call data extrinsic

related to https://github.com/ChainSafe/Multix/issues/80


packages/ui/src/types.ts
```diff
@@ -1,3 +1,7 @@
+import { GenericCall } from '@polkadot/types'
+
```
Tbaut, line 2:
```suggestion
```

asnaith:
Working well with the call data I've tried. Just found some minor issues (which could be tackled in other issues let me know). 

- The string "From Call data" has capitalization inconsistency (other options only capitalize first word)

- No error for bad call data just a "."

Screenshot 2023-06-16 at 12 49 19 PM

- There's an introduction of a layout issue with two elements being too close together :

https://github.com/ChainSafe/Multix/assets/3982665/de427521-6ba9-4332-ada9-6211f3e5eb96

asnaith (approved):
Fixes are good to go :)

PR #217, opened 2023-06-19
Remove the batch and more info

Tbaut:
- remove the batch extrinsic
- I think it makes sense to keep the manual extrinsic. After all it does help for things like `proxy.removeProxy` etc.
- added some info box on the manual to direct users to "From call data" with a direct link. I hope they read it :shrug: 

image


asnaith (approved):
Agreed, it makes sense to keep the manual extrinsic so that users can get themselves out of other situations (eg like mistakenly rejecting the request to remove an older multi sig and wanting to re-initiate that process etc). 

Good call on adding the info box too!

PR #225, opened 2023-06-22
Better onboading experience

Tbaut:
- disable the create button and add loader upon signing the multisig creation
image

- show a watched account if the extension is connected but no account is shared and the watched account has multisigs
- move from h1 (super big) to h3  error message if no account connected and no watched account
- In case your first account ever was created, you'd have the waiting screen, and the info alert afterward still. This fixes it.

### DIFF
--- a/packages/ui/src/components/NewMulisigAlert.tsx
+++ b/packages/ui/src/components/NewMulisigAlert.tsx
@@ -32,6 +32,7 @@ const NewMulisigAlert = ({ className = '', onClose }: Props) => {
 
 export default styled(NewMulisigAlert)`
   width: 100%;
+  margin-top: 1rem;
 
   .infoText {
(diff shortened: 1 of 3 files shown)


asnaith:
These are really nice improvements 🙌

Only thing to point is that now that we're now only directed to the screen once the multisig is ready it makes this message redundant as it's already by the time we get to this screen: 

Screenshot 2023-06-23 at 10 26 17 AM

asnaith:
Please disregard the comment above, some misunderstanding on my behalf. 

That banner is only displayed if the user is already watching an account and it disappears when the newly created multisig is available which is to be expected. Users that are not watching an account will not see the banner and only end up on the home screen once everything is available, also expected. 

Good to go :)

asnaith (approved):

PR #229, opened 2023-07-02
Bump deps

Tbaut:
- update all deps
- keep `@subsquid/typeorm-store` at the current major version as bumping isn't supported for fire squid (what we use)
- add graphql resolution to avoid an error with different graphql versions
- remove older resolutions for `@graphql-typed-document-node/core`


asnaith (approved):
I ran through a full multisig / pure setup plus balance transfer on this branch. Looks good to go.

Issue #132, opened 2023-05-10
Improve the ux of modal dismissal 


asnaith:
Modals are nicely dismissed when the user clicks outside of their display area but with batch transfers, this now increases the risk of causing user frustration if they accidentally dismiss a window they have populated with data. 

We should explore ways that protect the user from accidentally dismissing. 

This follows from a comment on a PR. Original comment: 

_Whilst crafting a transaction I had to switch between different apps and another browser to copy addresses. A few times when returning to the browser with multix, I clicked the window to activate it and accidentally clicked out of the modal which lost everything I was setting up for the batch transaction. Perhaps on this modal, we should have a confirmation in this context before closing the dialog? This has always been dismissable like this but could now potentially be a cause for user frustration if they were to lose long batched params that they've been in the middle of crafting._  

_Originally posted by @asnaith in https://github.com/ChainSafe/Multix/issues/114#issuecomment-1541013982_

Tbaut:
I really really dislike popup asking "Are you sure blablabla" and I think this shows that something has been badly thought.
If we think that users don't land on a modal by mistake, meaning that they understood what they were clicking on to land here. Then it's very unlikely that they want to leave right away, as a result we can have the close behavior a little more complex to achieve than just a click outside. For consistency in the UI, we need all modals to behave the same as much as possible.

Therefor, I suggest that we don't close on click outset, and have either
- a close button
- a close cross at the top left

I'm thinking the first one is probably nicer, but the second one works better on mobile.

Any opinion?

asnaith:
> I'm thinking the first one is probably nicer, but the second one works better on mobile.
> 
> Any opinion?

Agreed. A pop-up would look messy because the modal would already be stacked and another pop-up stacked on top of that, happy to avoid that!

I think a manual close option would work perfectly. As for which type would be best, we should probably choose the nicest solution for where we expect the majority of users to connect from and go for the close button if expecting most people to be on Desktop.

Tbaut:
Hey @dj-pundir, yes feel free to work on this. Just to be clear, all the modals should have a cross in the top right hand corner to close it and the behavior to be closed on click should be removed.

PR #230, opened 2023-07-03
Add modal dismissal

dj-pundir:
This is the resolution of issue https://github.com/ChainSafe/Multix/issues/132

Followings are the changes:
- Add close icon for dialog dismissal.
- Prevent dialog dismiss on user outside click.
- Use Close Icon button from the existing page "NewMulisigAlert"
- Add dismissal on all dialogs existing currently in the application.
- Run the lint --fix command.

Here is the popup dialog now:

image


asnaith:
Just an FYI @Tbaut, the CloudFlare deploy to preview failed here due to an api key issue

asnaith:
Thank you for the contribution @dj-pundir! This is working nicely. 

I'll wait for @Tbaut to approve as I think he will want to review the code before it's merged and right now we're only requiring one approval.

Tbaut:
Yes cloudflare deployment doesn't like external contributions unfortunately. 
Sorry for the long wait for review, I've been on some big PRs lately. Will take a closer look tomorrow.

packages/ui/src/components/modals/ChangeMultisig.tsx
```diff
@@ -342,10 +351,22 @@ const ChangeMultisig = ({ onClose, className }: Props) => {
...
+            className="closeButton"
+            size="small"
+            aria-label="close"
+            color="inherit"
+            onClick={onClose}
+          >
+            <CloseIcon fontSize="small" />
+          </IconButton>
+        ) : null}
+      </DialogTitle>
```
Tbaut, line 369: I suggest to place it as the first element, outside of the `DialogTitle` and also no need to check for the `onClose`, this will never be undefined.

```suggestion
        <IconButton
          className="closeButton"
          size="small"
          aria-label="close"
          color="inherit"
          onClick={onClose}
        >
          <CloseIcon fontSize="small" />
        </IconButton>
      <DialogTitle>Change multisig</DialogTitle>
```

Tbaut (changes requested):
Thanks a lot @dj-pundir for your contribution. I suggested a little change (please change it for all the occurences). The rest looks good.

packages/ui/src/components/modals/ChangeMultisig.tsx
dj-pundir: Done, Thanks!!

packages/ui/src/components/modals/ChangeMultisig.tsx
```diff
@@ -356,9 +365,17 @@ const ChangeMultisig = ({ onClose, className }: Props) => {
...
     >
+      <IconButton
+        className="closeButton"
+        size="small"
+        aria-label="close"
+        color="inherit"
+        onClick={onClose}
+      >
+        <CloseIcon fontSize="small" />
+      </IconButton>
```
Tbaut, line 378: Add the check on `isCallStep` to keep the previous behavior
```suggestion
      {!isCallStep && (
        <IconButton
          className="closeButton"
          size="small"
          aria-label="close"
          color="inherit"
          onClick={onClose}
        >
          <CloseIcon fontSize="small" />
        </IconButton>
      )}
```

Tbaut (changes requested):

packages/ui/src/components/modals/EditNames.tsx
```diff
@@ -130,4 +139,10 @@ export default styled(EditNames)`
     margin-bottom: 1rem;
     align-items: end;
   }
+
+  .closeButton {
+    position: absolute;
+    right: 0.5rem;
+    top: 0.5rem;
```
Lykhoyda, line 146: We are getting rid of classes in our code. Please use the styled component to apply the styles (https://emotion.sh/docs/styled). As you are using the same icon and styles 3 times, you might create a component for that, f.e in src/components/library/CloseButton.tsx

Tbaut (approved):
Thanks a lot @dj-pundir I think this is good to go.

PR #239, opened 2023-07-04
Use subsquid aquarium

### LINKED ISSUE #224: Deploy networks on subsquid Aquarium
The current bill for our db and indexer is pretty high. We should experiment using Subsquid aquariums

Tbaut:
closes #224 

- moves squid outside the monorepo because Aquarium only supports npm build
- change indexer to store accounts by chain (chain-id) to use only 1 DB for all networks
- adapt the UI to these changes
- ci deployment (no reset) on any change to squid. Reset needs to be done manually.

How to review:
Nothing should have changed if compared to the live Multix
- connect to any network beside Rhala (self hosted, not changed)
- any multisig/proxy for owned/watched accounts should be displayed
- there should be no UI difference between what's live and this PR's UI deployment


.github/workflows/deploy-squid-stage.yaml
```diff
@@ -1,121 +1,31 @@
...
-env:
-  # secrets are listed here to simplify the readability
-  AWS_REGION: '${{ secrets.AWS_REGION }}'
-  AWS_ACCOUNT_ID: '${{ secrets.AWS_ACCOUNT_ID }}'
-  AWS_ROLE_NAME: '${{ secrets.AWS_ROLE_NAME }}'
-  APP_STAGE: 'stage'
-  DOCKER_REGISTRY: 'ghcr.io'
-  DOCKER_TAG: 'latest'
-  ENV_FILE: '${{ secrets.ENV_FILE }}'
+      - tbaut-use-aquarium # run this workflow if the branch starts with ci/*
```
mpetrunic, line 11: comment doesn't match value

.github/workflows/lint-squid.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
       - name: install packages
-        run: corepack enable && yarn install --immutable
+        run: npm install
```
mpetrunic, line 26:
```suggestion
        run: npm ci
```
to use lockfile

.github/workflows/lint-squid.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
       - name: install packages
-        run: corepack enable && yarn install --immutable
```
mpetrunic, line 26: cannot comment above but instead of L20 we should use setup-node and provider nodejs version and cache

packages/ui/src/constants.ts
```diff
@@ -21,48 +22,55 @@ export const PAYMENT_INFO_ACCOUNT = '5CXQZrh1MSgnGGCdJu3tqvRfCv7t5iQXGGV9UKotrbf
 
 export const networkList = {
   kusama: {
+    chainId: 'kusama',
     explorerNetworkName: 'kusama',
     rpcUrl: 'wss://kusama-rpc.polkadot.io',
-    wsGraphqlUrl: 'wss://kusama.multix.chainsafe.io/graphql',
-    httpGraphqlUrl: 'https://kusama.multix.chainsafe.io/graphql',
+    wsGraphqlUrl: 'wss://squid.subsquid.io/multix-test/graphql',
```
mpetrunic, line 28: aren't we suppose to use versioned urls?

squid/yarn.lock
mpetrunic, line 1: seems like squid is using npm so this is not needed^^

.github/workflows/deploy-squid-stage.yaml
Tbaut: right, I should remove this line alltogether before merging. Will change the comment in the mean time
```suggestion
      - tbaut-use-aquarium # run this workflow if the branch starts is tbaut-use-aquarium
```

packages/ui/src/constants.ts
Tbaut: This is the prod url. If we bump the `squid.yaml` manifest version, and run the `sqd deploy` a new version will be created with the version in the url. This can then be tested thoroughly, and upon clicking "make prod" in the Aquarium UI, the versioned api endpoint is directed to this endpoint, with 0 downtime. I should change the name of the project, in order to have just `multix` though, and not "test" in the name

.github/workflows/lint-squid.yaml
Tbaut: sounds good, I added npm cache in [4a55656](https://github.com/ChainSafe/Multix/pull/239/commits/4a5565682508f5484a03a9361569effc1a757a0f)

mpetrunic (approved):

asnaith:
I did some smoke testing on this branch. 

✅ Create a new multisig wallet/contract.
✅ Attempt to send funds from a multisig wallet and encounter rejection.
✅ Successfully send funds from a pure proxy.
✅ Propose a valid transaction with specific call data and have it accepted.
✅ Propose an invalid transaction with call data, accept it, and see the error.
✅ Change the threshold for a multisig setup.
✅ Add a new member to an existing multisig wallet.

asnaith (approved):
Looks good to go

PR #249, opened 2023-07-12
Close button modal nits

Tbaut:
- rename the close button to ModalCloseButton because it's specifically styled.
- remove the close button when there's an error in the multisig change. To see this, make the same Tx twice to change a multisig (signatory rotation)
- delete stray console.log


packages/ui/src/components/library/ModalCloseButton.tsx
```diff
@@ -3,20 +3,22 @@ import { Close as CloseIcon } from '@mui/icons-material'
 
 interface Props {
   onClose: () => void
+  className: string
```
Tbaut, line 6:
```suggestion
  class Name?: string
```

asnaith (approved):

PR #250, opened 2023-07-13
Show hash and optional call data input

### LINKED ISSUE #245: Show nice hash and allow users to paste the callData for verification
Follow up on https://github.com/ChainSafe/Multix/pull/244

- Allow users to paste the callData
- mention that we're showing the call hash
- better styling

Tbaut:
closes #245 

- For a Multix Tx, show the hash and decode the call
image

- For a non Multix tx, show the hash and the call data field. The field is optional in case this is not the latest call (e.g in a 3/3 threshold, for the 2nd approval, the callData isn't actually needed, but it's good to have for users to verify).

Optional
image

Required
image


asnaith (approved):
Oh nice to have these options!

- Validated the process of constructing a call in the Polkadot JS extrinsic tab, obtaining the encoded call data, and submitting the proposal through Multix for multisig approval requiring no further posting of call data other than from the proposer.

- Validated the process of constructing a call in the Polkadot JS extrinsic tab, submitting the proposal in Polkadot JS, and ensuring that in a 2/2 threshold, the subsequent signatory opening the proposal in Multix must provide the call data for finalization.

- Validate the process of constructing a call in the Polkadot JS extrinsic tab, submitting the proposal in Polkadot JS, and verifying that in a 3/3 threshold, the subsequent signatory opening the proposal in Multix has the option to provide call data to enhance the experience for the final signer, but if not provided, the last signatory must supply it.

(Added the above to the test plan too)

Tbaut:
Just a nit @asnaith 
> the subsequent signatory opening the proposal in Multix has the option to provide call data to enhance the experience for the final signer

The option to provide the calldata is only to decode the call and verify it, it will not be submitted on chain, and the last signatory will have to provide it as a requirement. I can't go around this, it's either the first signatory using Multix, or the last. The others can't provide it on-chain (I mean they can, but this won't help). That's a limitation of the way the multisig pallet is built.

asnaith:
@ tbaut Thanks for the clarity, I've updated the test plan accordingly

PR #251, opened 2023-07-13
Add asset hubs, astar, hydra

Tbaut:
Adds support for:
- Asset hub Polkadot
- Asset hub Kusama
- HydraDx
- Astar


asnaith:
- If the previously selected network name was shorter than the maximum there's a slight discrepancy: 

Screenshot 2023-07-13 at 5 25 33 PM

- I wasn't able to connect to the Hydra node

Tbaut:
Any suggestion for the network selector? Asset hub is the new name for statemine/statemint and although I don't want to keep this long overflowed name, I didn't come up with something better yet.

The networks are all live but I forgot to switch to the versionned URL, my bad. Let me fix that and give you a couple addresses to watch maybe.

Tbaut:
For Hydra, you can watch Leemo that's part of a multisig: 7Hsq5RH9xUtPWFZMGXtoVWNd4CEjpJWsidf7bcGwNwdxp9Ha
And I found a multisig with a pending call that can be decoded `7LUnmQPXGrZ2AzVp5ccuK6apg95rxDxYnxKmCgHUZtDzUf9a`

Otherwise I can give you ways to find multisigs in DM.

asnaith:
> 

Oh I meant for Hydra it would not actually connect to the URL (an onfinality domain) but now I see that it's connecting to the node at wss://hydradx-rpc.dwellir.com and is working fine and was able to verify with Leemo's address.

asnaith (approved):

PR #255, opened 2023-07-14
fix border-radius

### LINKED ISSUE #253: Unify border radius
Some input elements don't have the same border radius, e.g in the send modal.
Note that this is one instance of such an inconsistency, please take a look around to find these and fix them all in the same PR.

image

Lykhoyda:
Closes #253

### DIFF
--- a/packages/ui/src/components/Transactions/Transaction.tsx
+++ b/packages/ui/src/components/Transactions/Transaction.tsx
@@ -141,7 +141,7 @@ export default styled(Transaction)(
     max-height: 1.3125rem;
     left: 29px;
     top: 19px;
-    border-radius: 0 0.9375rem 0.9375rem 0.9375rem;
+    border-radius: ${theme.custom.borderRadius};
     padding: 0.25rem 0.5rem;
     max-width: 2.625rem;
     font-size: 0.625rem;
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -141,7 +141,7 @@ export default styled(Transaction)(
     max-height: 1.3125rem;
     left: 29px;
     top: 19px;
-    border-radius: 0 0.9375rem 0.9375rem 0.9375rem;
+    border-radius: ${theme.custom.borderRadius};
```
Tbaut, line 144: This breaks the effect of having the top left not having any in the top left. Not sure if this suggestion works, we we could try

```suggestion
    border-radius: 0  ${theme.custom.borderRadius}  ${theme.custom.borderRadius} ${theme.custom.borderRadius};
```

Tbaut (commented):
Thanks, it def looks more consistent that way.
There's just a little thing that I made a suggestion for. If that work I think it's better, if not we can forget as this will be changed eventually.

packages/ui/src/components/Transactions/Transaction.tsx
Lykhoyda: I think the new design doesn't have that view for the badge. So I think we can consider it as a deprecated and keep the new design

Tbaut: this is not implementing the new design though, we're in the old, changing border-radius. Now if that's something you've strong opinion about, because you don't like it, I can let it go.

Tbaut (approved):

Tbaut:
I guess it's ready to merge?

Lykhoyda:
> I guess it's ready to merge?

Now we can merge, got the reply from Cindy about unified border-radius

PR #257, opened 2023-07-17
Fix label display

### LINKED ISSUE #252: Add signing label back
To be more informative about the action user is supposed to do, we need to add back the label "Signing with"

image

Lykhoyda:
Closes #252


Tbaut (approved):
awesome, thanks

Issue #242, opened 2023-07-10
Use `GenericAccountSelection` in lieu of `Autocomplete` when possible


Tbaut:
`packages/ui/src/components/selectors/AccountSelection.tsx` makes use of an `Autocomplete` with custom logic that could be replaced by the shared `GenericAccountSelection`.

I personally didn't research this deeply but I can't think of any reason why this shouldn't be possible with minor changes

PR #261, opened 2023-07-17
Fix icon and lack of badges

### LINKED ISSUE #254: Icon too small and lack of badge
The icon in the drop down here is too small, and these accounts should display a badge at least in the list (pure or multi)
image

Tbaut:
closes #254 
mostly address https://github.com/ChainSafe/Multix/issues/242 with `SignerSelection` now using the shared logic
removes quite a some hacks introduced in https://github.com/ChainSafe/Multix/pull/235/files#diff-1fd9795facc51dab124c0e926c52c9de7943bc717c296edc1454a7a9550fb7da

todo: the `AccountSelection` component somehow doesn't pick up the styling from the `Autocomplete`. My brain is fried, I'm sure you'll know what's wrong @Lykhoyda


Lykhoyda:
@Tbaut it doesn't pick up the styles because you removed this line:
```
const appGlobalStyles = css`
  .MuiAutocomplete-popper {
    margin-top: 0.75rem !important;
    ${AutocompleteStyles}
  }
`
```

When we render something in the "portal", it's rendered outside the root div. So it's not possible to access the styles without making them global.

Tbaut:
Why would all the other drop-down have the new style though?

Lykhoyda:
> Why would all the other drop-down have the new style though?

All others are rendered without a portal. We are using the portal only for the case with a long list to render the options on top of other components.

here is the difference that makes it happen in BalancesTransfer
```
disablePortal={false}
```

```
<GenericAccountSelection
        className="to"
        accountList={acountBase}
        onChange={onAddressDestChange}
        value={selected}
        label="to"
        disablePortal={false}
        allowAnyAddressInput={true}
      />
```

Tbaut:
Ah I see thanks for your input. I'll find a way then.

packages/ui/src/App.tsx
```diff
@@ -1,32 +1,23 @@
...
-  }
-`
 
 const App = () => {
   const queryClient = new QueryClient()
 
   return (
     <MuiThemeProvider theme={theme}>
       <CssBaseline />
-      <GlobalStyles styles={appGlobalStyles} />
```
Lykhoyda, line 29: As described in the comments without global styles for Autocomplete the feature disablePortal={false} will not work. It helps us to render big lists and it's related to the following discussion https://github.com/ChainSafe/Multix/pull/235#discussion_r1258931751

packages/ui/src/components/IdenticonBadge.tsx
```diff
@@ -3,18 +3,28 @@ import { useMemo } from 'react'
...
 import { AccountBadge } from '../types'
 import MultixIdenticon from './MultixIdenticon'
+import { ICON_SIZE_MEDIUM, ICON_SIZE_SMALL } from '../constants'
 
 interface Props {
   className?: string
   badge?: AccountBadge
   address?: string
+  sideBadge?: boolean
+  small?: boolean
```
Lykhoyda, line 13: I would prefer to have iconSize props with the string to set the size in rem. Or to stick with other components and have a prop 
```
sizeVariant?: 'small' | 'medium' | 'large' 
```

Lykhoyda: Also, Icon sizes are more situational rather than unified, so putting a strict value is not a fit for every situation

Lykhoyda:
There is no gap between the input and options list, but I think it's related to a similar issue with the global styles

Screenshot 2023-07-18 at 16 15 09

packages/ui/src/App.tsx
Tbaut: Thanks for your help. I got rid of `disablePortal` altogether, keeping the standard way.

packages/ui/src/components/library/Autocomplete.tsx
```diff
@@ -101,15 +96,9 @@ const Autocomplete = <
   )
 }
 
-Autocomplete.defaultProps = {
-  iconSize: '1.5rem',
-  disablePortal: false
```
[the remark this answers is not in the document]
Tbaut: It shouldn't be relevant any more, I got rid of both these props.

Tbaut:
@Lykhoyda 
> There is no gap between the input and options list, but I think it's related to a similar issue with the global styles

Yes, the styles were not picked up, because it's an `AccountSelection`. That's solved.

packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -165,5 +165,7 @@ export default styled(AccountDisplay)`
 
   .identityBadge {
```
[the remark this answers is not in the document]
Tbaut: I didn't touch this part, and would rather keep this PR focused on the main thing, fixing the issue.

packages/ui/src/components/IdenticonBadge.tsx
Tbaut: done

packages/ui/src/components/library/Autocomplete.tsx
```diff
@@ -96,42 +92,19 @@ const Autocomplete = <
         renderOption={renderOption}
         popupIcon={<HiOutlineChevronDown />}
         renderInput={renderInput}
+        PopperComponent={PopperMy}
```
Lykhoyda, line 95: @Tbaut Nice catch with the Popper component!

Lykhoyda, line 95: We can pass PopperStyled directly without the function wrapper 
```
PopperComponent={PopperStyled} 
```

Tbaut: working nicely :+1:

Tbaut:
I've switch from margin to padding, and 0.25rem instead of 0.75, also added a padding-bottom so that the space was consitent whether it opens from the top of the bottom. I think it looks good like this

image
image

Just the network drop down is now inconsistent. But that's not introduced here and not the subject of this PR either, so I'll open an issue.

asnaith:
I noticed the multi-badge is being displayed alongside the identicon in the drop-down rather than in the top left and the option is bigger than the value in the dropdown (I'm assuming this is the same styling thing Anton already mentioned). 

The account address in the option is also being truncated when there's room to display it, this is not an introduction though. Separate issue or quick enough to fix here?

Screenshot 2023-07-18 at 3 11 39 PM

Tbaut:
No, Anton mentioned a bug that I had to fix (and that I mentioned in the todo in the description, that I removed when fixed)

What you mention is all intended and part of the figma design actually. Is is something that you find not looking good or weird?

Lykhoyda (approved):

asnaith:
> No, Anton mentioned a bug that I had to fix (and that I mentioned in the todo in the description, that I removed when fixed that's why you missed it)
> 
> What you mention is all intended and part of the figma design actually. Is it something that you find not looking good or weird?
> 
> Here some examples, every time we want a smaller icon, we put the badge on the side. e.g (note that this particular screen isn't consistent but you get the idea.) image

@Tbaut Yeah, I guess it does seem a little weird to me within the drop-downs tbh but probably subjective. It looks fine in the transaction queue! So long as it's intentional, all good with that just wanted to ensure that was the case.

PR #263, opened 2023-07-18
update icons; integrate react-icons lib

### LINKED ISSUE #195: Use HeroIcons across the app
we should use `HeroIcon` as introduced here https://github.com/ChainSafe/Multix/pull/189 for all the icons.
- Search for `@mui/icons-material` and replace.

Lykhoyda:
Closes #195


Lykhoyda:
@asnaith Please take a look if any icon looks weird or has an inappropriate size

packages/ui/src/components/Toasts/ToastContent.tsx
```diff
@@ -26,11 +23,11 @@ const ToastContent = ({ className, toast }: ToastContentProps) => {
...
       <div className="iconContainer">
         {type === 'success' ? (
-          <CheckCircleOutlineIcon className="toastIcon" />
+          <HiOutlineCheckCircle size={20} />
         ) : type === 'loading' ? (
-          <CircularProgress className="toastIcon" />
+          <CircularProgress size={20} />
         ) : (
-          <ErrorOutlineIcon className="toastIcon errorIcon" />
+          <ErrorOutlineIconStyled size={20} />
```
Tbaut, line 30: let's do size 25 here so that they are aligned with the text. 20 is a bit small.
image

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -53,7 +54,10 @@ const Transaction = ({
         {!aggregatedData.callData ? (
           <QuestionMarkIcon className="callIcon unknownCall" />
         ) : (
-          <GestureIcon className="callIcon" />
+          <GestureIcon
```
Tbaut, line 57: There is something wrong with this one, or with the changes done in this file
image

Tbaut (commented):
Thanks a lot, I went and check for each, and found some small issues.

packages/ui/src/components/Transactions/Transaction.tsx
```diff
@@ -129,8 +129,7 @@ export default styled(Transaction)(
   }
 
   .callIcon {
-    font-size: 7rem;
```
Tbaut, line 132: Here, the font size is the issue, if I put back this line, I get what I expect
image

asnaith:
> @asnaith Please take a look if any icon looks weird or has an inappropriate size

I inspected all screens and modals and they look good to me. The only icon I noticed had changed was the three-dot options button! The only size issue I noticed was the toast icon that was already mentioned.

Lykhoyda:
@Tbaut thank you for the review. I updated the code

Tbaut (approved):
Looking great :rocket:

PR #265, opened 2023-07-19
Show more info with needed and reserved funds

### LINKED ISSUE #248: More information regarding the deposit and fees when sending a tx
> Would be cool to have Multix get the current deposit value and tell you you're too poor by X amount

This is the feedback from someone I helped figuring out why the send modal what displaying the following on Polkadot:
image

BTW Polkadot needs 20 DOT deposit for every multisig proposal (you get it back :|)

Tbaut:
closes #248 
To test, create a MSig with as signatories at least 1 account that you control but doesn't have funds. When you selected it to submit a TX it won't have enough funds (even on Rococo where the fees are ridiculous)
I tested on Polkadot too
image
image


asnaith (approved):
Working great and a really nice improvement 🚀

Issue #268, opened 2023-07-20
Optimize squid queries


mpetrunic:
**Description**
Graphql query to fetch all multisigs is quite complex and generates very unoptimized SQL query. To make things worse, when we do subscription, it continuously executes that unoptimized SQL query every 5 seconds. We haven't noticed that before because every network was in it's own database schema which meant there was a lot less rows in tables.
Current grapql query: https://github.com/ChainSafe/Multix/blob/86c486945c2ac4223ab682c68bc0204fd09f43fc/packages/ui/src/queries/multisigsByAccounts.graphql

**Prerequisites**

Regular usage

**Steps To Reproduce**

Open multix

**Expected Behaviour**

Super fast response and optimized query

**Potential Solutions**

### 1. Separate everything into different databases.
 - Pro:
    - "simple" solution that will buy us some time
 - Con:
    - once aquarium introduces billing it will be expensive for us
    - we need to change frontend to query different endpoints for each network

### 2. Optimize query

The following query generates two very optimized and fast sql queries which allows us to get multisig accounts. For delagator/delagatee stuff we would need to jump on a call with whiteboard to challge what we actually need.

```graphql
query MultisigsByAccounts($accountIds: [String!], $watchedAccountIds: [String!]) {
  accountMultisigs(where: { signatory: { id_in: $accountIds } }) {
    multisig {
      address,
      isMultisig,
      isPureProxy,
      signatories {
        id
      }

    }
  }
  accounts(where: {id_in: $watchedAccountIds     AND: { isMultisig_eq: true, OR: { isPureProxy_eq: true } }}) {
      address,
      isMultisig,
      isPureProxy,
      signatories {
        id
      }
  }
}
```

Note: for subscription, we would need to send that as two separate subscription querys.

### 3. Keep same db but generate models for each network

For example on squid side, instead of:
```graphql
type Account @entity {
  id: ID! @index @unique
  address: String!
  multisigs: [AccountMultisig] @derivedFrom(field: "signatory")
  isPureProxy: Boolean @index
  # it's the origin/delegator for another proxy accounts, a pure must have some
  delegatorFor: [ProxyAccount!] @derivedFrom(field: "delegator")
  # it's the delegatee/doing stuff on behalf of another account
  delegateeFor: [ProxyAccount!] @derivedFrom(field: "delegatee")
  isMultisig: Boolean @index
  signatories: [AccountMultisig!] @derivedFrom(field: "multisig")
  threshold: Int
[...]

PR #269, opened 2023-07-20
Move main networks to separate squids

Tbaut:
part of #268


asnaith:
I'm unsure on how to verify this, I can connect to each network but do we need to view a msig or pure within each one?

Tbaut:
To test you can just connect to the networks yes.. and try to see the diff in timing with prod for loading the multisigs..

asnaith (approved):

PR #270, opened 2023-07-20
Move last component to use `GenericAccountSelection` and unify address selection

### LINKED ISSUE #242: Use `GenericAccountSelection` in lieu of `Autocomplete` when possible

Tbaut:
closes #242 
- this unifies the design and code for all the address/signatories/account selectors
- fix an issue where the address was also bold (font-weight 500) e.g
image

The biggest part of the refactoring (and things to test) is regarding the new signatory selection in new multisig creation, as well as the field to add watched accounts.

### DIFF
--- a/.github/workflows/deploy-squid-aquarium.yaml
+++ b/.github/workflows/deploy-squid-aquarium.yaml
@@ -12,6 +12,9 @@ on:
 
 jobs:
   deploy:
+    strategy:
+      matrix:
+        network: ['polkadot', 'kusama', 'rococo', 'assethub-kusama', 'assethub-polkadot', 'others']
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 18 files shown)


asnaith (changes requested):
-  The input clear button is missing / not displayed. 

- It seems that the space for the identicon is now being reserved which means when you click an empty input the cursor does not start fully left aligned. It looks a bit strange and and only makes sense once the account is fully selected and the identicon is displayed. 

- In prod the user will not see the network prefixed address if they paste their SS58 address, it would just convert to the network specific one and display that after it's been added. Now when you paste an SS58 it shows the network-specific account address as a drop-down option, was this intentional? I think it makes sense for people who are already familiar with polkadot addressing (the core of users at the moment) but could be confusing for less familiar users.   

Screenshot 2023-07-20 at 2 33 04 PM

asnaith (approved):

Lykhoyda (approved):

PR #271, opened 2023-07-20
Hot fix - option button for multisig without pure proxy

Tbaut:
I don't remember how we broke this, but it sounds old.
We were not showing the the option menu for multisigs without proxy.

What I tested:
- watched accounts (multisig or proxy) have no send button
- watched accounts (multisig or proxy) have an option menu to change names
- owned accounts (multisig or proxy) have a send button


asnaith (approved):
Good to go :) 

An example of something to automate in the near future.

Tbaut:
> A watched multisig-only was not showing the options button in prod but this branch has fixed that.

Yup sorry, this is the fix actually. I'll edit the title

PR #273, opened 2023-07-25
New design for pure proxy header

### LINKED ISSUE #259: New Design for pure proxy view
We want to come close to this:
image

Figma: https://www.figma.com/file/4dQC9kxf5uxCrAjpkzAK0D/Multix?node-id=839%3A27723&mode=dev

- This view is only visible when there is a pure proxy
- we can forget about the circle around the identicon
- if there is no pure proxy, the multisig will be displayed and there will be no such header to start with. See https://github.com/ChainSafe/Multix/issues/167

Lykhoyda:
Closes #259 

The Home page was refactored with the State-Container and Presentation components pattern. That the home page is responsible for displaying the main elements f.e. the grid and importing the View components as PureProxyHeaderView. The Home component is responsible for the data which should be shared across other presentation components. All the UI components and styles were moved to the individual components.

In the end, we have 3 individual components to keep code corresponding with the single responsibility principle:
- MultisigView
- MultisigAcftionMenu
- PureProxyHeaderView

Also, I refactored Modal's state management by creating ModalsContext to avoid the props drilling in HomePage and children components. The functionality wasn't changed, the refactoring goal was to improve the readability of the code and enhance the code structure to be able to change the code without influencing other components on the Home page.
Screenshot 2023-08-15 at 17 31 51


Tbaut:
Thanks, I'll check it out when I'm back. The overview is a link to the Help > overview section

asnaith:
Hey @Lykhoyda, I found a couple of things but overall it's looking great! 

- At the moment, we are showing the "New Transaction" button for non-connected, watch-only accounts (causes an application crash when the button is clicked).

- On devices with a width lower than 378 pixels (iPhone 11 and below) the text from the button overflows. 
Screenshot 2023-07-25 at 12 59 53 PM

- On mobile views, when the width is so narrow that the balance of the pure needs to be displayed underneath the identicon, it would be good to have some padding between the elements. 
Screenshot 2023-07-25 at 12 55 50 PM

asnaith (changes requested):
Hey @Lykhoyda, I found a couple of things but overall it's looking great! 

- At the moment, we are showing the "New Transaction" button for non-connected, watch-only accounts (causes an application crash when the button is clicked).

- On devices with a width lower than 378 pixels (iPhone 11 and below) the text from the button overflows. 
Screenshot 2023-07-25 at 12 59 53 PM

- On mobile views, when the width is so narrow that the balance of the pure needs to be displayed underneath the identicon, it would be good to have some padding between the elements. 
Screenshot 2023-07-25 at 12 55 50 PM

Lykhoyda:
@asnaith I didn't check the mobile version properly, thank you for the review! I updated the code, should be better

asnaith:
The latest changes for the mobile layout look great @Lykhoyda 🙏

I think you may have missed the first bullet point, regarding the "New Transaction" button appearing for watched pures (should only be the overflow button if they are not a signatory and just viewing). 

https://github.com/ChainSafe/Multix/assets/3982665/ec0eb246-a4fa-49b2-b356-c1c0a5d06b94

Lykhoyda:
@asnaith Indeed I missed the first bullet, but thank you for the video it is clear what the problem is. I uploaded the fix. Please share your Proxy address so that I can test the watch function.

asnaith:
> @asnaith Indeed I missed the first bullet, but thank you for the video it is clear what the problem is. I uploaded the fix. Please share your Proxy address so that I can test the watch function.

@Lykhoyda I was viewing 5EYcEL7mAjuXfZGhEEutoKTMwkBj9xFHkju43kpf8BfuKntL 

It's working great with the latest changes! 🚀

asnaith (approved):

asnaith:
Ah actually I just noticed the balance position can change, not sure if this is because of the pure having an identity or because the new transaction button is not shown when in watch mode (for example watch DCZyhphXsRLcW84G9WmWEXtAA8DKGtVGSFZLJYty8Ajjyfa on Kusama). Could be dealt with in a separate issue I guess? 

Screenshot 2023-07-27 at 10 37 59 PM

packages/ui/src/components/IdenticonBadge.tsx
```diff
@@ -90,7 +95,14 @@ const BadgeStyled = styled(Badge)<{ badgeType: AccountBadge; sideBadge: boolean
         position: relative;
         transform: unset;
       `}
-  }
+
+    ${({ size }) =>
```
Tbaut, line 99: Can we also have the "pure" written a little bit bigger
image
the figma has `font-size: 0.96431rem; `

packages/ui/src/pages/Home/LoaderWrapper.tsx
```diff
@@ -0,0 +1,25 @@
+import { Box, Grid } from '@mui/material'
+import { styled } from '@mui/material/styles'
+import React, { PropsWithChildren } from 'react'
+
+const LoaderWrapper = (props: PropsWithChildren) => {
```
Tbaut, line 5: please add some comment in the PR description to what brought you to add this.

packages/ui/src/pages/Home/MultisigActionMenu.tsx
```diff
@@ -0,0 +1,54 @@
+import { Button } from '../../components/library'
```
Tbaut, line 1: Same here, please let us know if anything has changed. It's most probably just bringing this logic to a new component with copy/pasting, yet the fact that I can't properly check what has changed may introduce bugs that go unnoticed until we're in prob.

packages/ui/src/pages/Home/PureProxyHeaderView.tsx
```diff
@@ -0,0 +1,160 @@
...
+  background: ${({ theme }) => theme.custom.gray[300]};
+  border-top-right-radius: ${({ theme }) => theme.custom.borderRadius};
+  border-top-left-radius: ${({ theme }) => theme.custom.borderRadius};
+`
+
+const PureHeaderStyled = styled('div')`
+  display: flex;
+  flex-wrap: wrap;
+  margin: 0 0 1rem 0.5rem;
+
```
Tbaut, line 70:
```suggestion
```

Tbaut (commented):
Thanks a lot for this. This looks really good!!

Given the large amount of change, I'd appreciate a proper PR title and description explaining what has changed, and the additions compared to what the issue contains. E.g what `loaderwrapper` brings, or tell Andrew what components have changed slightly so that he can test or visually check.

Any thought on what we could do better for the mobile, view only button to look better? Maybe aligning to the right?
image

packages/ui/src/pages/Home/LoaderWrapper.tsx
mpetrunic: khm code comment khm 😛

packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -99,10 +102,12 @@ const AccountDisplay = ({
               />
             )}
             {!!sub && <span>{sub}</span>}
-            <NameStyled>{localName || mainDisplay}</NameStyled>
+            <NameStyled className={'multisig__name'}>{localName || mainDisplay}</NameStyled>
```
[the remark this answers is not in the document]
Lykhoyda: BEM is always in my mind :)

packages/ui/src/components/IdenticonBadge.tsx
Lykhoyda: done

asnaith:
Latest changes look good to me 🤩

Lykhoyda:
> Thanks a lot for this. This looks really good!!
> 
> Given the large amount of change, I'd appreciate a proper PR title and description explaining what has changed, and the additions compared to what the issue contains. E.g what `loaderwrapper` brings, or tell Andrew what components have changed slightly so that he can test or visually check.
> 
> Any thought on what we could do better for the mobile, view only button to look better? Maybe aligning to the right? image

For now, I show the "Settings" word instead of vertical dots on the mobile version as it would be easier to click and understand due to the button size

packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -99,10 +102,12 @@ const AccountDisplay = ({
               />
             )}
             {!!sub && <span>{sub}</span>}
-            <NameStyled>{localName || mainDisplay}</NameStyled>
+            <NameStyled className="multisigName">{localName || mainDisplay}</NameStyled>
```
Tbaut, line 105: This class name is used and set in another component `PureProxyHeaderView`. There's the chance that we change/remove the other component without realizing we actually change this one. Can we centralize the design here?

packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -99,10 +102,12 @@ const AccountDisplay = ({
...
           </NameWrapperStyled>
         )}
-        <AddressStyled>{getDisplayAddress(encodedAddress)}</AddressStyled>
+        <AddressStyled className="multisigAddress">
```
Tbaut, line 108: Same comment here.

Tbaut (commented):
Just found a small thing. If you can do something that'd be awesome @Lykhoyda and we can merge right after :clap:

packages/ui/src/components/AccountDisplay.tsx
Lykhoyda: Account Display is a generic component used in the other components and looks the same. For PureProxyHeaderView we need style adjustments. The properties are deeply nested so it's hard to target Name and Address, for that reason I created a class to change the styles. It will have no impact to other components if you remove PureProxyHeaderView

Tbaut: Right, but if you remove the classe names in `AccountDisplay.tsx`, which is something we've done before, because we said that we should avoid the use of classNames, then it'll break. We've had the issue in the past, and I'd like to avoid it as much as possible. However, I do definitely get your point, and making things super complex.

Maybe adding a comment here would help, like "className used here for specific cases in the AccounHeader for instance". What do you think?

Lykhoyda: For me, it looks like a regular thing and it's easy to detect if your style is off as you have visual feedback. But if you find this helpful, I added the comments in Account Display

Tbaut: Thanks. I don't like that kind of comment either, but again, it's because we've broken this in the past. I can't find the PR now, but there was this case, where, for cleaning, we removed some `classNames`, and migrated them to styled components, and things just broke all over without being spotted in review, because it's super hard to spot.

Issue #205, opened 2023-06-13
Possible EventEmitter memory leak detected


asnaith:
**Description**
After some time of my browser being in the background I went to use Multix and noticed a generic error being displayed on the main screen. 

After viewing the console it implies there is a "Possible EventEmitter memory leak detected" 

**Steps To Reproduce**

Use multix and then leave it backgrounded, return to it an hour or so later (not 100% sure about these steps yet). 

**Expected Behaviour**
Multix should display as normal 

**Console Error**

<img width="1093" alt="Screenshot 2023-06-13 at 12 49 50 PM" src="https://github.com/ChainSafe/Multix/assets/3982665/31479b83-4157-4867-8fa0-44940a554b03">

Full console dump:
[console_log.txt](https://github.com/ChainSafe/Multix/files/11738847/console_log.txt)

**Screenshots / Video**
<img width="777" alt="Screenshot 2023-06-13 at 12 50 54 PM" src="https://github.com/ChainSafe/Multix/assets/3982665/70d7c0fb-ddce-49cd-8dfa-24034de24497">

**Test Host Specifics:**

- OS: macOS
- Firefox 

Not sure if this issue is OS or browser dependent but I've only seen it happen with this combo at the moment. 

**Additional context**
- As soon as I refreshed the browser there was no longer an issue so this is not hugely user-impacting

Tbaut:
Thanks, this looks like disconnections issues that aren't handled well by our `useSubscription`. We have 2 long lasting connections to the indexer, and both seem to timeout from time to time. Note that the ws connection to the node made by the api is also timing out, but probably handled better and not crashing the app.

Tbaut:
I have encountered infinite loading of the multisigs. I have another idea, will try it out in the next days.

PR #277, opened 2023-07-31
Add retry subscription with error message

Tbaut:
Attempt at closing #205
Adding some more logs and see if this works.


asnaith (approved):
Looks like you've solved the issue :) 🚀

I left three different browsers open with Multix tabs, and resumed over 1.5 hours later without seeing the error that we used to see before.

Tbaut:
I couldn't manage to have the error again either, but I'm not 100% convinced this is what actually solved it tbh, because it's supposed to print a specific console error when retrying.. and I didn't see it. Maybe it's also related to the dedicated subsquid processors that we now have. Anyway I'll merge, the issue will stay open, and we can see with time how it behaves.

PR #279, opened 2023-08-01
Use optimized queries

### LINKED ISSUE #205: Possible EventEmitter memory leak detected

### LINKED ISSUE #268: Optimize squid queries

Tbaut:
- closes #268 
- closes #205 by implementing a `refetch` on error. 
- Set the pure to appear first in the list.

This implements the 2nd suggestion to have optimized queries to subsquid.
It adds a nice to have that if you add a watch account that's a signatory for 1 multisig, with 1 proxy above, you'll end up actually seeing in the Multix all the other multsigs that the proxy has below it. It wasn't the case before.

eg, on Kusama `Alice u Bob` with address `DGaHm71gRHJGQVF2ubunkx37qgLpno2gYibq2WHoDJpFMt6` is only part of 1 multisig, that has the ChaosDAO pure above `DCZyhphXsRLcW84G9WmWEXtAA8DKGtVGSFZLJYty8Ajjyfa`. If you add this former account as watched, you'll see the ChaosDAO pure along with all the other multisigs below it as if you were watching the pure.

Regarding testing, this is reworking the whole multisig queries and logic so the best way to test is:
- using watch account, know what to expect and add an account. Verify that the proxy/multisigs that you see are the ones expected.
- using your own accounts, verify that the dropdown displays the same multisigs/pure as on prod.
- I launched a huge subsquid like the one that was very slow and this branch is deployed against it. we should try to see if things are slow at times.


Tbaut:
I launched a huge squid like the one that was very slow the past week and this branch is deployed against it. we should try to see if things are slow at times.

asnaith:
🟢 I can now see that all the other multisigs under the pure, not just the multisig the watched signatory is part of
🟢 When switching between networks all resolved within 2 - 5 seconds (saw the same on prod too)
🟢 Verified that my connected accounts/dropdown looked the same as prod
🟢 I see the pure as the first option in the dropdown list now. However....
🟠 When we add a signatory to watch, the multisig is still the default selection in the dropdown not the pure. Is this expected?

Tbaut:
About your last point, it's indeed expected. We have no logic to automatically switch to a given account because
- how should we chose one if there are many, and how should the user know that there are many
- this logic would definitely be a little hacky, since the whole UI is reactive and just adapts to new changes automatically.

The best way I can think of to handle this would be something like
- make a dedicated 1 time query to see if there are pure/msig linked to this account
- inform the user (in a modal?) that there are now x new accounts to explore in the drop-down, we could offer a way to switch to them with 1 click.

Tbaut:
FYI @asnaith I saw this time my fix working for the re-connection to the indexer after a while. The error message to look for is "Subscription [some id] error, refetching"

packages/ui/src/contexts/MultiProxyContext.tsx
```diff
@@ -64,136 +76,116 @@ const MultiProxyContextProvider = ({ children }: MultisigContextProps) => {
...
+                  signatories: getSignatoriesFromAccount(multisig.signatories),
+                  threshold: multisig.threshold
+                }
+              ]
+            } as MultiProxy)
+          }
+        })
+
+        // flatten out multisigMap
+        const multisigArray = Array.from(multisigMap.entries()).map(([, multip]) => multip)
```
mpetrunic, line 117:
```suggestion
        const multisigArray = Array.from(multisigMap.values())
```
wouldn't that do same thing without allocating array twice? :)

packages/ui/src/contexts/MultiProxyContext.tsx
```diff
@@ -64,136 +76,116 @@ const MultiProxyContextProvider = ({ children }: MultisigContextProps) => {
...
+        // find the delegatee that are multisigs and put all the infos right away
+        account.delegatorFor.forEach(({ delegatee, type }) => {
+          if (delegatee.isMultisig) {
+            const previousMultisigsForProxy = pureProxyMap.get(account.address)?.multisigs || []
 
-        // if this account is a multisig and is the delegatee for at least a pureProxy
-        if (account.isMultisig && pureProxyAddresses?.length > 0) {
-          pureProxyAddresses.forEach(({ pureAddress, type }) => {
-            const previousMultisigsForProxy = pureProxyMap.get(pureAddress)?.multisigs || []
+            const isAlreadyInMultisigList = !!previousMultisigsForProxy.find(
```
mpetrunic, line 142:
```suggestion
            const isAlreadyInMultisigList = previousMultisigsForProxy.some(
```
More optimized as it returns immediately first time condition is true^^

packages/ui/src/contexts/MultiProxyContext.tsx
```diff
@@ -64,136 +76,116 @@ const MultiProxyContextProvider = ({ children }: MultisigContextProps) => {
...
-        ([proxy, agg]) =>
-          ({
-            proxy,
-            multisigs: agg.multisigs
-          } as MultiProxy)
-      )
-
-      res.push(...proxyArray)
+    // flatten out pureProxyMap
+    const pureProxyArray = Array.from(pureProxyMap.entries()).map(
```
mpetrunic, line 165: again .values() will save on memory allocation

Tbaut: absolutely

packages/ui/src/contexts/MultiProxyContext.tsx
Tbaut: nice catch, yes

packages/ui/src/contexts/MultiProxyContext.tsx
Tbaut: ah not for that one, we need both key and value here.

mpetrunic: ah true, but why do you need Array.from then?

Tbaut: the `.entries()` doesn't return an array, it returns an `IterableIterator<[string, Omit<MultiProxy, "proxy">]>` so you have to make this an array to be able to call `.map` on it.

mpetrunic (approved):

asnaith (approved):

Tbaut:
I'll keep the old indexers running for the next days, so that we can revert quickly if needed.

Issue #136, opened 2023-05-11
Support Walletconnect v2


Tbaut:
Walletconnect v2 is now supported in Nova. Users could connect their account and sign transaction using their phone if they can connect it to Multix.

Example of integration: https://github.com/paritytech/polkadot-staking-dashboard/pull/907/files

Tbaut:
ooh that's what we need: https://github.com/paritytech/polkadot-onboard

PR #282, opened 2023-08-07
Add modal context

Tbaut:
This is a small refactoring to set the stage for #136 
It's taking elements from #273 but only changes the logic of opening/closing the modals. Nothing should change beside the fact that it'll fix the issue that modals get closed if the list of multisig is refreshing.


asnaith:
Refactor looks good :) 

 No issues with the functionality or look of the edit, change, send modals.

asnaith (approved):

PR #285, opened 2023-08-10
Support `removeProxies` in the indexer

### LINKED ISSUE #283: Creating proxies + multisigs  and removing them afterwards
**Description**

Error found when trying to send tokens from pure proxy to multisig. Message will indicate "proxy not proxy"  as seen below. This is because there is actually no existing relationship between the pure proxy and the mutlisig in question. 

The indexer is watching for a specific event that happens when a proxy is removed. But the extrinsic _**removeProxies**_  which function, which essentially obliterates all connections, fails to emit this event.  The indexer doesn't pick it up and the indexer remains unaware and fails to detect the change. This makes it look like everything is the same and nothing has been removed. 

Screenshot 2023-08-08 at 9 17 54 AM

screenshot_2023-08-08_at_9 23 59_am

Tbaut:
closes #283 

This is non trivial to properly test, due to the way the indexer works. There are 3 types of scenarios:
1. it's a new event. Meaning that you call `proxy.removeProxies` now, on an old pure, that is already persisted in the DB
1. it's a recent event compare to the proxy creation, but we re-index the chain. Meaning that the `proxy.removeProxies` will be in a batch with potentially many other proxies creation/deletion
1. you create a proxy or a pure, then shortly after a `proxy.removeProxies`, and we re-index the chain (like on 2). Then the new proxy or the pure will never even touch the DB. The removal will happen in a batch array.

I've carefully tested all these scenarios. There is no easy way to test 2 and 3 because it requires a re-indexing, which happens every month, when we have changes in the indexer.

In any case the deployment of the test UI is now pointing to the new (v2) of the indexer that supports it. @LernaJ you should not see the pure proxies any more (the multisig are here still though)


asnaith (approved):
The **New Event** scenario is working well. The pure removal went smoothly, and the UI updated as expected without needing a refresh.

For the **Recent Event with Chain Re-Indexing** and **Concurrent Creation and Removal with Chain Re-Indexing scenarios**, I couldn't fully test them due to indexer knowledge / access but I think the testing you have already done is sufficient here.  It seems unlikely that critical proxies would be created and removed so rapidly in the real world.

Tbaut:
Thanks. @LernaJ  will be able to test the behavior with re-indexing because she did call the `proxy.removeProxies` in the past.
Also just to be clear, I can manually launch those re-index, so I could test all these scenarios locally.

Tbaut:
FWIW I hibernated the v2 to see if this is what's causing the slowness. As a result, this PR can't be tested any more, and obv shouldn't be merged before restarted.

Issue #286, opened 2023-08-10
Can't filter by name in the account selection


Tbaut:
Since https://github.com/ChainSafe/Multix/pull/270 we broke the filtering by name.
To reproduce:
- create a new multisig
- in the account selection, try to type a name.

-> the filtering doesn't work.

PR #287, opened 2023-08-10
Fix filtering by name

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/286


asnaith (approved):
CleanShot 2023-08-10 at 10 57 38@2x

Looks all good now. 

 (tbh I wasn't aware we did previously did this. I usually don't have too many connected at the same time and just select from the dropdown! )

Tbaut:
Hehe I do and realized pretty quickly! Also if we start having an address book it'll be handy.

PR #289, opened 2023-08-10
Fix nits on send field

### LINKED ISSUE #288: Balance transfer doesn't have a default value that makes sense
When sending funds with Send > Send tokens:
- the default value for the "to" field is our own account, which doesn't particularly make sense
- when clearing the field with the little "x", the identicon remains

Tbaut:
closes #288

### DIFF
--- a/packages/ui/src/components/EasySetup/BalancesTransfer.tsx
+++ b/packages/ui/src/components/EasySetup/BalancesTransfer.tsx
@@ -14,13 +14,12 @@ import { getOptionLabel } from '../../utils/getOptionLabel'
 interface Props {
   className?: string
   from: string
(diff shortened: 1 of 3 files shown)


asnaith:
@Tbaut I noticed the send field used to have name filtering on it but no longer does on this PR branch (was on my mind after testing the previous PR!). 

Other than that this looks good, does make a lot more sense to have an empty field and the clear button works well.

Tbaut:
This is actually as designed, there's no pre-selection, so there's no filtering. I made it work like the watched account field basically. But you're right, also having in mind a future address book, I guess we should keep the preselection. I'll add it tomorrow.

Tbaut:
Ok it was a quick one. I enabled auto-merge as I'm pretty confident.

asnaith (approved):

PR #290, opened 2023-08-15
Show multisig if they are delegatee for non pure proxies

Tbaut:
I was told on Hydra that this account was a multisig and not showing on Multix `7NJp5SL8d2SGhzwqdD7qnF4Z3VQe4uMefoxWFnQMuGJeLwPw`

I could verify and found the issue. The reason is that this account was delegatee for a non-pure proxy, and my logic had no case for it.


asnaith (approved):
Fix works great :)

PR #294, opened 2023-08-16
Add support for old anonymous proxies

### LINKED ISSUE #291: A pure created by a normal account then controlled by a multisig isn't displayed
We should understand and fix the reason why this account, `7PE5okM6CA7vGkpgHMd4hR6nqGddYySoqe6CAxKDgR4FAJoc`, a pure created here https://hydradx.subscan.io/extrinsic/2404978-2 isn't displayed.

Tbaut:
- closes #291 
- cleanup old config files now unused.

The issue was that the proxy pallet used to be called "Anonymous" and the indexer didn't support these events.
Hydra still used these for a while.

To test:
- watch 7PE5okM6CA7vGkpgHMd4hR6nqGddYySoqe6CAxKDgR4FAJoc
- see the pure and the multisig. This wasn't displayed in prod


asnaith (approved):
Showing up as expected now :) 

@Tbaut Is it "old" as in you can't create this type of pure anymore? Or do you just mean an older style (non-multix created) pure?

Tbaut:
yup, it's the old name given by an old version of the pallet. I didn't bother about it so far, but each parachain can chose to evolve at its own pace, and this user had created a pure proxy 4 months ago on hydra, where this old pallet version was live.

PR #297, opened 2023-08-17
Support Kusama earlier blocks

Tbaut:
I was asked to support Kusama for older blocks than 15M as we did now.
This requires to rework the logic I just did yesterday as we need to be runtime agnostic, what the current fire squid model doesn't allow. The Subsquid ppl told me they are re-working the type model in arrow squid (next version) and this will hopefully be better.

In any case we're now supporting block from Kusama 6M, and Hydra should also still have no issue.

Older accounts include:
Hydra: `7PE5okM6CA7vGkpgHMd4hR6nqGddYySoqe6CAxKDgR4FAJoc`
Kusama: `DYqDKe9uKGb9Zo68udDQ15CWE6b7Did3khGm6oYRy2TCCRn`


Tbaut:
Merging as there isn't much to test here.

PR #299, opened 2023-08-17
Change transaction's sorting order

### LINKED ISSUE #298: Sort by latest first
From a user on mobile: I takes me a lot of scrolling to get to the latest proposal to sign, and I'd prefer to have it sorted by latest first.

Tbaut:
closes #298


Lykhoyda (approved):

packages/ui/src/components/Transactions/TransactionList.tsx
```diff
@@ -160,11 +160,11 @@ const TransactionList = ({ className }: Props) => {
         const filtered = res.filter((agg) => agg !== undefined) as AggregatedData[]
         const timestampObj: AggGroupedByDate = {}
 
-        // sort by date
+        // sort by date, the newest first
         const sorted = filtered.sort((a, b) => {
           if (!a.timestamp || !b.timestamp) return 0
```
Lykhoyda, line 165: As a suggestion, I would move it to utils

Tbaut: Thank. While it's too specific for a shared util (it's dealing with `AggregatedData` which is defined here), I've defined this specific function outside of the component.

PR #310, opened 2023-08-18
Remove useless Wanring component

Tbaut:
Small refactoring as I had my eyes on it to remove a useless component.
This touches the following warning messages:
- change multisig with the pure not having enough funds
- any warning in the send modal (e.g not enough funds)
- account selection (e.g add account to a multisig that's already part of it)
- creation or multisig change when signatory doesn't have enough funds.


asnaith (approved):
I've confirmed that all the warnings are still working as intended after this change 👍

@Tbaut Thank you for the bullet points!

PR #311, opened 2023-08-18
Show a banner for our treasury proposal

### LINKED ISSUE #306: Add a banner to ask ppl to vote on our treasury proposal
Should be in the home, displayed to everyone with or without multisigs.
Redirecting to Polkassembly, Subsquare.

Closing it means we don't annoy users for this, ever (storing in localstorage).

Tbaut:
closes #306 

I went for just when there's something displayed in the home, and not when there's nothing to be found.
image


asnaith:
Great idea! 

We need a bit of padding between it and the pure header component. 

CleanShot 2023-08-18 at 11 13 47@2x

Tbaut:
ah, good one, I didn't check with a pure :pray: fixed

asnaith (approved):
Looks good to go now :)

PR #312, opened 2023-08-19
Fixed balance input in Manual Extrinsic

aadarsh012:
Previously, when you entered any balance param while making a Manual Extrinsic, the value that goes into the transaction differed greatly from the value entered, so I fixed it, now the right value goes into the txn.

You can see the difference in the attached screenshots and you can test this bug also in your product -> Go to Send -> Manual Extrinsic -> select pallet as 'balance' and method as 'transfer' (just for example) -> input values and make a transaction, now when the transaction appears in the ui, you can see the value is different.
Screenshot 2023-08-19 at 1 33 09 PM
Screenshot 2023-08-19 at 1 33 01 PM


Tbaut:
Thank you very much for this PR, I'd call this more an annoyance than a bug. The value is in plank and not in full tokens. But it's definitely better to have it this way :pray:

I'll take a deeper look in the next days. In the mean time I've got a couple comments. 

- please uncommit the changes in the `yarn.lock` as there are no dependamcy change.

-we should re-use the code from the "send-funds" for the input and show to users that we're now talking in Dot/Ksm... rather than in plank. This is extremely important since users may have been used to enter the plank amount.

aadarsh012:
Yarn.lock is fixed, no changes now.

Tbaut (changes requested):
You're actually introducing a bug here! 

Not only balances are compact `u16`, `u32`... Just have a look at `proxy.killPure` that accepts a `height` and a `extIndex`... well your code would multiply these it by the 10^(decimals)! I just check Polkasafe and yup you have this bug :)

Tbaut:
Can we expect this PR to be worked on further @aadarsh012 ?

PR #313, opened 2023-08-21
feat: update Multisig view with Accordion

Lykhoyda:


packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -1,61 +1,89 @@
...
-              <SignatoriesWrapperStyled>
-                <h4>
-                  Signatories{' '}
+            <Accordion
+              header={
+                <AccountDisplayWrapperStyled>
+                  <AccountDisplay
+                    address={multisig.address || ''}
+                    badge={AccountBadge.MULTI}
+                    withBalance={false}
```
Tbaut, line 25: I really think we should keep the balance visible at all time, and keep this.
image

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -1,61 +1,89 @@
...
+                  <BoxStyled>
+                    <Button variant="secondary">
+                      <Balance address={multisig.address} />
+                    </Button>
+                    {!selectedIsWatched && !selectedHasProxy && (
+                      <MultisigActionMenu withSettingsButton={false} />
+                    )}
+                  </BoxStyled>
+                  <div>
+                    <SignatoriesHeadingStyled>Signatories List</SignatoriesHeadingStyled>
```
Tbaut, line 41: Keeping it simple
```suggestion
                    <SignatoriesHeadingStyled>Signatories</SignatoriesHeadingStyled>
```

Lykhoyda:
@Tbaut  So the current to do that needs clarification (the rest is fixed):
- New transaction button position (sync the design)
- 3 dots button (not present in the design in the multisig view, but is present in the header view. The issue is that we don't show the Header view for standalone multisig)
- Always show a balance on multisig (imo it stretch  the multisig selector too much that it looks disproportionately )
- Make margin left on Signatories list (Probably better to clarify with designer)

packages/ui/src/pages/Home/MultisigView.tsx
Tbaut: Answering here regarding your comment:
> Always show a balance on multisig (imo it stretch the multisig selector too much that it looks disproportionately )

Agreed, maybe we should put it right to the Multisig, in its own box like with do with the header? But my point remains, we should see the balance without an additional click. Also I find it not nice right now in it own box below, it looks out of place.

@sweetpea22 help :D

Tbaut:
Your point 1 and 2 need to be discussed with @sweetpea22, let's have a meeting.

I answered the point 3 in the related comment above.

> Make margin left on Signatories list (Probably better to clarify with designer)

we can talk about it too with @sweetpea22 

You marked it as resolved, but I can still see the behavior from https://github.com/ChainSafe/Multix/pull/313#discussion_r1300688535

packages/ui/src/pages/Home/MultisigActionMenu.tsx
```diff
@@ -39,14 +43,14 @@ const MultisigActionMenu = ({ withSendButton = true }: MultisigActionMenuProps)
     <>
       {withSendButton && (
         <Button
-          variant="primary"
+          variant="secondary"
           aria-label="send"
           onClick={() => setIsSendModalOpen(true)}
         >
-          New transaction
+          New
```
Tbaut, line 50: I think the "new" is misleading a little since we don't know what we're talking about, is it a new multisig? And even more importantly, we need the 3 dots buttons too, because right now you have no ways to change the names if all you have is a multisig! I'd leave it primary button like for proxies, it's a little misplaced above everything, but as I suggested before, I think we can focus on make the header for solo multisigs next.

This is what we have.
image

previous ask, keeping it above, with primary is fine since it's the only one here.
image

Tbaut (changes requested):
This looks reaaaally good! I found a couple things, but nothing major.

packages/ui/src/pages/Home/MultisigAccordion.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { styled } from '@mui/material/styles'
+import React from 'react'
+import AccountDisplay from '../../components/AccountDisplay'
+import { MultisigAggregated } from '../../contexts/MultiProxyContext'
+
+interface AccordionProps {
+  multisig: MultisigAggregated
+}
+
+const MultisigAccordion = ({ multisig }: AccordionProps) => {
```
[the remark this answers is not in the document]
Lykhoyda: @Tbaut I updated it, it looks smoother now. But probably it's still far from perfect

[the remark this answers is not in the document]
Tbaut: with `exit: 0` it's actually behaving as I'd expect. See next comment.

packages/ui/src/pages/Home/MultisigAccordion.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return (
+    <AccordionMuiStyled TransitionProps={{ timeout: { appear: 300, enter: 300, exit: 200 } }}>
```
Tbaut, line 14:
```suggestion
    <AccordionMuiStyled TransitionProps={{ timeout: { appear: 300, enter: 300, exit: 0 } }}>
```

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -1,50 +1,60 @@
...
+          selectedMultiProxy.multisigs.map((multisig) => {
+            return (
+              <MultisigWrapperStyled key={multisig.address}>
+                <AccountDisplayWrapperStyled>
+                  <AccountDisplay
+                    address={multisig.address || ''}
+                    badge={AccountBadge.MULTI}
+                    withBalance={false}
+                  />
+                  {/*{ TODO: implement multisig transactions  */}
```
Tbaut, line 37: Should we remove these comments?

packages/ui/src/pages/Home/MultisigActionMenu.tsx
```diff
@@ -8,9 +8,13 @@ import { useModals } from '../../contexts/ModalsContext'
 
 interface MultisigActionMenuProps {
   withSendButton?: boolean
+  withSettingsButton?: boolean
```
Tbaut, line 11: This props is never used afaics, and doesn't have the right name anyways, these aren't settings. Should we remove this props altogether?

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -1,50 +1,60 @@
...
       <HeaderStyled>
         {selectedMultiProxy && (
-          <h3>{renderMultisigHeading(selectedMultiProxy.multisigs.length > 1)}</h3>
+          <h3>
+            {selectedHasProxy
+              ? 'Controlled by'
+              : renderMultisigHeading(selectedMultiProxy.multisigs.length > 1)}
+          </h3>
         )}
         <BoxStyled>{!selectedIsWatched && !selectedHasProxy && <MultisigActionMenu />}</BoxStyled>
```
Tbaut, line 24: We **must** still see the 3dots buttons for watched accounts. This dropdown is currently used for renaming, and isn't visible for watched multisigs. This is the case why the `withSendButton` prop was created.

Tbaut (commented):
awesome, we're getting there, but I still miss this 3 dots buttons for watched solo multitisgs (it's ok for proxies)

Tbaut (approved):
ok sweet, I pushed a change because the 3 dots button was there twice in the case of a proxy
image

Issue #308, opened 2023-08-18
Amend the main account dropdown behavior


Tbaut:
Right now, when we change networks, the Multiproxy selector is still visible until the new are set.
We should:
- clear the list of multiproxy on network switch
- show when we're looking for multiproxies (this is displayed on the home, but not in other pages)
- ~~redirect to the home when this resolves whether or not there's something in the list~~ edit: probably not a good idea

asnaith:
The multisigs / pures are being cleared and refreshed on network switch and the loading spinner is working nicely :) 

Just a couple of issues:

- Re-direct to home when the network resolves is not working for me (the issue implies this should be happening).

- Some nav links are being pushed underneath others

<img width="1279" alt="CleanShot 2023-08-28 at 13 18 30@2x" src="https://github.com/ChainSafe/Multix/assets/3982665/d520c01b-903d-48fb-aad7-95065cdae666">

PR #315, opened 2023-08-24
Don't show old accounts when loading a new network

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/308


asnaith:
The multisigs / pures are being cleared and refreshed on network switch :) 

Just a few of issues:

- Re-direct to home when the network resolves is not working for me (the issue implies this should be happening).
- There's no loading / progress spinner shown on mobile view, desktop is fine
- Some nav links are being pushed underneath others

CleanShot 2023-08-28 at 13 18 30@2x

asnaith (changes requested):
The multisigs / pures are being cleared and refreshed on network switch :) 

Just a few of issues:

- Re-direct to home when the network resolves is not working for me (the issue implies this should be happening).
- There's no loading / progress spinner shown on mobile view, desktop is fine
- Some nav links are being pushed underneath others
- Maybe another merge happened since this was submitted, there's now some component conflicts

CleanShot 2023-08-28 at 13 18 30@2x

Tbaut:
Thanks, I have no idea how this happened, or how I couldn't have seen it. In any case I fixed it. The loader was white and hence invisible on mobile, that's definitely from me :+1: 

Regarding the redirect to home, it comes from me originally, but now I wonder if this is a good idea. I think we should ideally not do any magic and redirect without a good reason. wdyt?

asnaith (approved):
Nav links and a spinner on mobile all look good with the latest changes :) 

Also indifferent regarding the re-direction so let's leave it for now, we can revisit it in the future if we're more convinced that there's a need.

PR #321, opened 2023-08-28
Docker for Multix development

ltfschoen:
* [X] Figure out how to resolve error `error from daemon in stream: Error grabbing logs: invalid character '\x00' looking for beginning of value` that is shown when viewing Docker logs by running `docker logs -f subsquid_db`
* [X] Use non-root user account Docker container
* [ ] Optional: Detect `NODE_ENV` and apply production settings (i.e. `yarn install --production`)
* [ ] Optional: Replace GraphQL URL that is used for `VITE_GRAPHQL_WS_PROVIDER` and `VITE_GRAPHQL_HTTP_PROVIDER` in ./packages/ui/.env and ./packages/ui/.env.example and `wsGraphqlUrl` with separate port (i.e. GQL_PORT), host, and proto, as shown here https://github.com/subsquid-labs/squid-substrate-template/blob/main/scripts/sub-client.js#L4.


docker/Dockerfile.dev
```diff
@@ -0,0 +1,79 @@
...
+
+# copy required dependencies first as they change less often in code
+COPY --chown=node:node ./packages/ui/package.json ${PATH_UI}/package.json
+# improves time to install packages
+# COPY --chown=node:node ./packages/ui/yarn.lock ${PATH_UI}/yarn.lock
+
+COPY --chown=node:node ./squid/package.json ${PATH_SQUID}/package.json
+# COPY --chown=node:node ./squid/package-lock.json ${PATH_SQUID}/package-lock.json*
+
+RUN apk update && apk add --update git vim sudo python3 make g++ && \
```
ltfschoen, line 41:
```suggestion
RUN apk update && apk add --no-cache --update git vim sudo python3 make g++ && \
```

Tbaut:
Hey @ltfschoen tbh I'm not sure we want to go this route, we used to have dockers for everything, but we figured this wasn't really useful and added quite some maintenance burden that we didn't want to have. Look for instance at https://github.com/ChainSafe/Multix/tree/7e0d01892ceb09654b6de66f6f2bf5159a32817c

ltfschoen:
> Hey @ltfschoen tbh I'm not sure we want to go this route, we used to have dockers for everything, but we figured this wasn't really useful and added quite some maintenance burden that we didn't want to have. Look for instance at https://github.com/ChainSafe/Multix/tree/7e0d01892ceb09654b6de66f6f2bf5159a32817c

that's ok no issues with it being closed, i usually do this anyway with new repositories that i try

PR #323, opened 2023-08-29
Refactor address encoding/decoding

### LINKED ISSUE #278: User can add an address to watchlist multiple times if different chain selected
**Description**

Watch mode has some error checking that prevents a user from adding the same account a second time. 

This can be circumvented by choosing a different chain other than the right one for the address. As the address is added it's converted for the other chain and hence allows you to add another. 

**Steps To Reproduce**

1. Have a pure address from one chain (eg Kusama) 
2. Add the Kusama Address to the watch list when Polkadot is the chain selected
3. Repeat steps 1 & 2 and see the same account can be added multiple times. 

**Expected Behaviour**
The user should only be able to add it once. I guess we need to check the converted address too. 

**Screenshots / Video**

https://github.com/ChainSafe/Multix/assets/3982665/0bd91a2a-22e6-47eb-bbcb-761b770f2dc3

Tbaut:
closes #278 

What started as a small pach, ended up being quite an important refactoring, with Moonbeam/river in mind (there's still the back-end to port). In any case, this is touching everything that is linked to an address. From the watch accounts, to the account displayed, the multisig creation/change

### DIFF
--- a/packages/ui/src/components/AccountDisplay.tsx
+++ b/packages/ui/src/components/AccountDisplay.tsx
@@ -9,7 +9,7 @@ import { useApi } from '../contexts/ApiContext'
 import { DeriveAccountInfo, DeriveAccountRegistration } from '@polkadot/api-derive/types'
 import IdentityIcon from './IdentityIcon'
 import { useGetBalance } from '../hooks/useGetBalance'
-import { encodeAddress } from '@polkadot/util-crypto'
+import { useGetEncodedAddress } from '../hooks/useGetEncodedAddress'
 
 interface Props {
   address: string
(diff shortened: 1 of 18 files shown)


asnaith (approved):
I did a test of all major functionality since this was extensive: 

- Create a new multisig wallet/contract.
- Attempt to send funds from a proxy and encounter rejection if balance is too low (balances.transferKeepAlive manual extrinsic).
- Successfully send funds from a proxy.
- Attempt to send funds from a multisig and encounter rejection if balance is too (balances.transferKeepAlive manual extrinsic).
- Successfully send funds from a multisig.
- Send a valid transaction with specific call data and have it accepted (System.remark manual extrinsic).
- Reject a transaction successfully
- Send an invalid transaction with call data,  acceptance, and see an error.
- Send a valid transaction using call data, accept and see successful execution
- Change the threshold for a multisig setup.
- Add a new member to an existing multisig wallet.
- Watch an account

All was expected so we're good to go

Tbaut:
Wow thanks a lot for all these tests, certainly things I haven't tested myself in a while :)

PR #326, opened 2023-08-30
Support WalletConnect v2 as a wallet

### LINKED ISSUE #136: Support Walletconnect v2

Tbaut:
closes #136 
todo:
fix the fixme

How to test.. fasten your seatbelt, it'll be bumpy.
- ssh with local port forward into our hydra rococo subsquid arquive `ssh -L 7777:localhost:7777 root@[redacted]` and leave this terminal open, we won't do anything with it, it's relaying what we need to your local 7777.
- locally in `/squid` make sure that your env is
  ```bash
  # hydra rococo
  RPC_WS="wss://hydradx-rococo-rpc.play.hydration.cloud"
  CHAIN_ID='hydra-rococo'
  BLOCK_START=1617000
  PREFIX=63
  ARCHIVE_URL="http://localhost:7777/graphql"
  ```
- launch the indexer based on this archive with this `docker compose down; docker compose up db -d; npm run codegen; npm run typegen; npm run build; npm run db:migrate; node -r dotenv/config lib/main dotenv_config_path=.env`
- in another terminal launch the graphql server to server this to the UI `npm run start:graphql-server`
- locally in `/packages/ui` make sure that your .env is
  ```bash
  VITE_WALLETCONNECT_PROJECT_ID=[redacted]
  VITE_CHAIN_ID="hydra-rococo"
  VITE_NETWORK_NAME="hydra-rococo"
  VITE_WS_PROVIDER="wss://hydradx-rococo-rpc.play.hydration.cloud"
  VITE_GRAPHQL_WS_PROVIDER="ws://localhost:4350/graphql"
  VITE_GRAPHQL_HTTP_PROVIDER="http://localhost:4350/graphql"
  ```
- run `yarn && yarn start`
- connect to the `local` network in the dropdown. Create a multisig.
- to test, go to https://rococo-app.hydradx.io/trade > connect wallet > Walletconnect. Copy the key by clicking below the QR code
- in Multix > settings > WalletConnect paste this key, and connect.


asnaith:
I appreciate the clear setup instructions!

I'm having a few issues though and have been unable to see any WC signing modal. This is what I'm seeing atm: 

**Local Storage and Wallet Connect:**

After inputting the Wallet Connect key and approving the connection, I see these keys in local storage

CleanShot 2023-08-30 at 11 56 37@2x

They do not update if I click the "Approve" button. They are already there prior, after connecting. 

Also, once connected, options like the "Create One" button or "New Multisig" link are not visible.
Even if I manually navigate to the create screen, I can't create a multisig because an error message appears, stating, "At least one of your accounts must be a signatory." 

**Main Connect Button Issues:**
- The main connect button in the top navigation bar persists even after a successful Wallet Connect connection.

**Settings' Connect Button:**
- The connect button under settings should be deactivated until a key is entered. Additionally, it would be helpful to indicate an active Wallet Connect connection here and provide an option to disconnect.

**Watched Address Component Display:**
- The watched address component is not rendering properly on the screen.

CleanShot 2023-08-30 at 10 46 53@2x

**Issues with the 'WallectConnect Connection Request' Modal:**
- The "Reject" button does not function as expected. The modal remains on screen despite the console displaying "Emitting history_updated."

- If the modal is closed, attempting to reconnect shows an error that a pairing already exists. This can be fixed by clearing local storage, but perhaps the storage should automatically clear upon rejecting or closing the modal?

asnaith (changes requested):
I appreciate the clear setup instructions!

I'm having a few issues though and have been unable to see any WC signing modal. This is what I'm seeing atm: 

**Local Storage and Wallet Connect:**

After inputting the Wallet Connect key and approving the connection, I see these keys in local storage

CleanShot 2023-08-30 at 11 56 37@2x

They do not update if I click the "Approve" button. They are already there prior, after connecting. 

**Cannot create a new multisig**

Once connected, options like the "Create One" button or "New Multisig" link are not visible.
Even if I manually navigate to the create screen and paste in two addresses (one of them being the account used to get the WC key) I can't create a multisig because an error message appears, stating, "At least one of your accounts must be a signatory." 

**Main Connect Button Issues:**
- The main connect button in the top navigation bar persists even after a successful Wallet Connect connection.

**Settings' Connect Button:**
- The connect button under settings should be deactivated until a key is entered. Additionally, it would be helpful to indicate an active Wallet Connect connection here and provide an option to disconnect.

**Watched Address Component Display:**
- The watched address component is not rendering properly on the screen.

CleanShot 2023-08-30 at 10 46 53@2x

**Issues with the 'WallectConnect Connection Request' Modal:**
- The "Reject" button does not function as expected. The modal remains on screen despite the console displaying "Emitting history_updated."

- If the modal is closed, attempting to reconnect shows an error that a pairing already exists. This can be fixed by clearing local storage, but perhaps the storage should automatically clear upon rejecting or closing the modal?

Tbaut:
Thanks for checking it out:  regarding the section "Local Storage and Wallet Connect" I think it's not relevant, this is a walletconnect part that is managed by the library and that I don't have my hands on.
"Cannot create a new multisig" and "Main Connect Button Issues:" sections are also expected. This feature is to connect the Multisig accounts to any Dapp. It is not meant to replace the connection from your accounts to Multix. 

- What is missing, is that it should not allow you to connect a Dapp, if you don't have any multisig/proxy that you hold a key for
> The connect button under settings should be deactivated until a key is entered. Additionally, it would be helpful to indicate an active Wallet Connect connection here and provide an option to disconnect.

Noted for the disabled button. Regarding the active connection, they are visible when effective. Just something went wrong because you didn't have any multsig to share I guess.

> The "Reject" button does not function as expected. The modal remains on screen despite the console displaying "Emitting history_updated."

I'll try it out, but i suspect there's again something unintended here because of no owned account.

> If the modal is closed, attempting to reconnect shows an error that a pairing already exists. This can be fixed by clearing local storage, but perhaps the storage should automatically clear upon rejecting or closing the modal?

That also unfortunately not something I have my hands on. The way to fix this is to re-create a key from the dapp, eventually by refreshing it.

asnaith:
Ok pushing the reset button now misunderstandings are cleared up 😅

I was able to successfully establish a Wallet Connect session once I updated my Vite config to use the right port. Deleting the session worked as intended too. 

To simplify things please disregard everything mentioned prior, here are the remaining things I think we need to look at: 

- I think we should disable the wallet connect field from being entered until there's a connected extension, and the wallet connect button should remain disabled unless someone has entered a key (This would stop someone from getting into the confusing state I was in earlier 😄)

- It would be good to have unique button names for connecting the extension and for connecting the wallet (with both being "Connect" I thought they did the same thing). Maybe "Create Session"? 

- The watched address component is not rendering properly on the screen

- The reject button does not work as intended (checked again, even with an owned account)

Tbaut:
Thanks for the thorough tests
- called the button "Connect Dapp", hoping it helps
- disable if the currently selected multiproxy is a watched one with a warning
- fixed the scrollbars, only visible on Brave.. that's a common FF vs BRave thing I forgot, good that we test on different browsers
- closing the modal on session rejection now.

I'll try to dig into the session already proposed modal.

asnaith:
Nice, I was able to establish wallet connect sessions connected to both pure or multisig and successfully perform TX (swaps on Hydra) using each. Previous issues resolved too :) 

Anecdotally, the experience for signing swaps is a little weird on Hydra (although Multix side is working as I would expect). As soon as the first (initiating) signatory has signed in Multix, the modal on Hydra reverts to the default layout. The tx for the 2nd approval goes into the Multix queue as expected and when signed the swap is executed but there's no reflection of this in Hydra due to the way their dapp handles things once the first signing is complete. Not really our issue but worth mentioning to them if we have a contact there. Perhaps they need to add additional logic on their dapp side now that multisig swaps are possible. Here's a video to show what I mean:

https://github.com/ChainSafe/Multix/assets/3982665/f8e6b8dc-eaa9-4e15-a024-21fec78ee857

---

Remaining Issues: 

- The reject button doesn't work on "WalletConnect Transaction signing" modal

- I was able to delete a wc session many times but did run into an issue where I could not

https://github.com/ChainSafe/Multix/assets/3982665/d1ec199e-bba3-407f-b5e0-2b4f559d4e1e

[cannot_disconnect_log.txt](https://github.com/ChainSafe/Multix/files/12491296/cannot_disconnect_log.txt)

Tbaut:
> As soon as the first (initiating) signatory has signed in Multix, the modal on Hydra reverts to the default layout.

> Not really our issue but worth mentioning to them if we have a contact there. Perhaps they need to add additional logic on their dapp side now that multisig swaps are possible. 

Dapps using WalletConnect do not have in mind that their Dapp will be used with multisigs. Adding some logic there would be super hacky. Their first use case is mobile wallet users, and I don't think this is scalable to try to make them change. As such, all we can do is educate our users, that they are using WalletConnect in a somewhat unconventional way, and that they should be ready for such behaviors.

Tbaut:
- fixed the reject signing button that didn't close the modal

> I was able to delete a wc session many times but did run into an issue where I could not

Despite trying hard, connecting/disconnecting, refreshing both UIs, I didn't manage to reproduce this issue.

An earlier issue that we had:
> If the modal [from a session request] is closed, attempting to reconnect shows an error that a pairing already exists. 

I made it so that closing the modal is the same as rejecting the connection request. Note that this is not the same behavior as with signing. With signing, if you close the modal, and refresh the page, the signing request will appear again. I chose to keep it as is because upon entering the connection key, the modal appears directly, while with a signing request, users will be on the Dapp rather than Multix, so they may come later, close the modal first to do something else... The signing will eventually expire if they don't sign/reject within a couple minutes.

asnaith:
> I fixed the reject signing button that didn't close the modal.

Confirmed. Reject button on "WalletConnect Transaction signing" modal is now good 👍

> Despite trying hard, connecting/disconnecting, refreshing both UIs, I didn't manage to reproduce this issue.

I have also not been able to reproduce the "unable to disconnect" issue so perhaps something was wrong with my local setup. As it's not prevalent and reproducible lets continue moving forward and if we see it again / can reproduce it we can tackle it separately (plus the user can always disconnect on the connected dApp too). 

> I made it so that closing the modal is the same as rejecting the connection request. 

This approach makes sense but at the moment I'm not seeing the change reflected in Multix, neither seems properly rejected (reject button or closing the modal) although the modal disappears. 

https://github.com/ChainSafe/Multix/assets/3982665/c647d27a-3e06-45ad-bf30-206ef0047c8d

I think now only the above and the FIXME stuff is remaining for this.

Tbaut:
The todos don't need to be made in this PR, those should be/are additional issues since this feature won't be much visible unless they dig.

Regarding the rejection. Since the user effectively rejected the proposal (your previous comment mentionned explicitely the fact that they closed the modal). Hydra was still showing the same QR code. Now that we reject the proposal, they can go to Hydra, that will not present the QR code any more (because of the rejection) and if they request a new key from hydra, then they get a new valid one.

asnaith:
> The todos don't need to be made in this PR, those should be/are additional issues since this feature won't be much visible unless they dig.
> 
> Regarding the rejection. Since the user effectively rejected the proposal (your previous comment mentionned explicitely the fact that they closed the modal). Hydra was still showing the same QR code. Now that we reject the proposal, they can go to Hydra, that will not present the QR code any more (because of the rejection) and if they request a new key from hydra, then they get a new valid one.
> 
> edit: ah yes, the FIXME was a workaround the fact that hydra seems to have their genesis hash hardcoded.. so the hydra rococo didn't actually show the right hash, it showed the live one, so the UI prevented us from doing anything because it wasn't matching. I'll fix the FIXME before merging and test against hydra.

Ok sounds good and I've just seen the behaviour in Hydra with the dialog closing etc. Makes sense to leave as is. 

So looks like we can merge as soon as the workaround for Hydra Rococo is removed :)

asnaith:
> I did get an approve, but not a formal one. Removed the fixme, and tested with the prod hydra without issues 🎉

Yep, can confirm here too - all good with this merge! :) 🚀

Issue #276, opened 2023-07-31
Support parachains that don't have the proxy pallet


Tbaut:
While unlikely, some parachain don't have the proxy pallet. cf #262 
We should rework the multisig creation to have the proxy creation optional, ideally check by default, and unchecked, with explanations for parachains without the proxy pallet.

PR #328, opened 2023-08-31
Support chains without a proxy

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/276

We now add a checkbox in the multisig creation flow for user to choose wether they want a proxy or not. It's checked by default, and won't be visible for chains that don't have the proxy pallet.
Because there's actually nothing to do to create such a simple multisig, what we do is propose to create a system remark, which is super cheap and contains "Multix creation [address]". We need to have at least 1 on-chain transaction from the multisig in order for the indexer to pick it up and know about this new multisig.

Whether this transaction is approved or not is actually irrelevant.. it just shows users what it's like to have a multisig transaction pending then :man_shrugging:


asnaith (approved):
I could successfully create multisigs with or without a pure after this. Looks good :)

PR #332, opened 2023-08-31
Filter if proxy isnt the selected one

### LINKED ISSUE #316: if the tx is proxy.proxy only show those related to the currently selected one
How to test:
- 1 multisig, 2 proxies.
- make 1 transaction involving any of the proxies
- this TX is visible in both proxies but should ideally not

Tbaut:
closes #316


Tbaut:
I'm self confident on this one, it's small and is requested in prod.

PR #334, opened 2023-09-04
Support Moonbeam/river

### LINKED ISSUE #333: Support Moonbeam/Moonriver
self explanatory. Note that although they have the proxy pallet, it's deactivated.

Tbaut:
closes #333

### How to test.. fasten your seatbelt, it'll be bumpy bis.
- ssh with local port forward into our moonbase subsquid arquive `ssh -L 7777:localhost:7777 root@[redacted]` and leave this terminal open, we won't do anything with it, it's relaying what we need to your local 7777.
- locally in `/squid` make sure that your `.env` file is
```bash
# moonbase
BLOCK_START=5043000
PREFIX=1284
RPC_WS="wss://wss.api.moonbase.moonbeam.network"
ARCHIVE_NAME="moonbase"
CHAIN_ID='moonbase'
IS_ETHEREUM='true'
ARCHIVE_URL="http://localhost:7777/graphql"
```
- launch the indexer based on this archive with this `docker compose down; docker compose up db -d; npm run codegen; npm run typegen; npm run build; npm run db:migrate; node -r dotenv/config lib/main dotenv_config_path=.env`
- in another terminal launch the graphql server to server this to the UI `npm run start:graphql-server`
- locally in `/packages/ui` make sure that your `.env` is
```bash
VITE_CHAIN_ID="moonbase"
VITE_NETWORK_NAME="moonbase"
VITE_WS_PROVIDER="wss://wss.api.moonbase.moonbeam.network"
VITE_GRAPHQL_WS_PROVIDER="ws://localhost:4350/graphql"
VITE_GRAPHQL_HTTP_PROVIDER="http://localhost:4350/graphql"
```
- run `yarn && yarn start`
- connect to the `local` network in the dropdown. You're now connected to Moonbase.
- You'll need some ethereum accounts, e.g using Talisman and connecting them there.
- You'll need some DEV tokens, ping me, and I'll send you a bunch
- The proxy pallet is deactivated, so you can't create a proxy there.


asnaith:
Will look at this as soon as https://github.com/ChainSafe/Multix/pull/326 is merged and this branch is updated with that code :)

asnaith:
1) I was able to encounter a SenderInSignatories Error. This was because I could select the account from the connected extension and also paste in the same address, this only happens with EVM addresses, an error is displayed if trying to do the same with a Polkadot account. 
(https://moonbase.subscan.io/extrinsic/0x8123a1344dbd4ccadba6f760e990e39c9b0cccbe083a7c7577f67abd229a3a07). 

https://github.com/ChainSafe/Multix/assets/3982665/16b9bc6c-40b3-44ec-a5b6-95ac552627f8

2) The fee calculation for the multisig creation seems off (maybe because no proxy?). If I connect accounts with no balance I see that "The selected signer requires at least 2.01651 DEV". However, I noticed when funding and signing it is only 0.000014 DEV. 

3) I was able to initiate a multisig creation with one account but I can't sign with the other two accounts as the transaction queue never loads. I'm just remaining on the "loading your multisigs" spinner.  This is stopping me from testing further. (https://moonbase.subscan.io/extrinsic/0x14f0538b91cc3342a6df1a23836f885425977127760c8757b3ade5ed168cb0d3)

asnaith:
Fixes for fee and adding same EVM account are looking good 🔥

**Actions checked with Moonbase EVM accounts**

🟢 Created a moonbase multisig successfully, remark tx showed as expected
🟢 Checked minimum balance warnings
🟢 Checked multisig balance is displaying / deducted properly
🟢 Checked Subscan links to Moonbase
🟢 Checked EVM address can be copied from the identity badge
🟢 Checked overview works with EVM addresses
🟢 Checked transaction signing modals display correct info and balance
🟢 Checked an amount can be sent successfully from the multisig

I thought it would be best to check the pre-existing stuff too so..

**Actions checked with regular accounts on Rococo:**

🟢 Multisig + pure creation
🟢 Balance transfer + approvals
🟢 Change amount of signatories 

I've only found two more things 

1) The EVM address icon does not show if a non EVM network is selected, expected I guess but connecting EVM accounts to other networks and leads to another point 2. 

CleanShot 2023-09-13 at 12 22 47@2x

2) There's a crash when connected to another network eg Rococo with EVM account and you type a number for the threshold, see video below: 

https://github.com/ChainSafe/Multix/assets/3982665/752e5307-bd13-45fd-96bc-330e269452db

Tbaut:
great find. I've fixed both issues. One very edge case is where:
- you are on an Ethereum network, you add eth addresses to the signatories
- you switch network for a non ethereum network
- we won't allow you to go further without a warning. 

This is such an edge case that I chose to stop making the code more complex at the cost of a questionable UX for a very edge case that may never happen.

asnaith:
> This is such an edge case that I chose to stop making the code more complex at the cost of a questionable UX for a very edge case that may never happen.

Yeah, agree with your assessment 👍

Tbaut:
Thank you so much for the super in depth tests

PR #337, opened 2023-09-05
Add Cypress

### LINKED ISSUE #335: Add Cypress e2e testing base code
We want to use Cypress to test Multix.
This issue is to do the base work
- add Cypress and its deps to Multix
- add the GH action running the tests for each commit
- add a simple test to watch an account on Rococo

Tbaut:
closes #335


Tbaut:
I figured @Lykhoyda that with the yarn bin, there's no need of `corepack enable` or anything along those lines. Since I couldn't figure out a way to make it work with yarn 2. Cypress supports it right of the bat, just it didn't catch it so far, up until the moment the bin was present. https://github.com/cypress-io/github-action#yarn-modern

asnaith:
Excellent! Checked it out, looking great. Thank you 🙌

I have a few small tweaks I'd like to make here and there but just nits. LGTM though, I can just add upon it after and make this first test more exhaustive etc 

🚀🚀🚀

asnaith (approved):

Tbaut:
I didn't spend much time on the test, and already ran into weird issues like the text I typed being cut.. Need to investigate.

.github/workflows/cypress-tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+  pull_request:
+    paths:
+      - 'packages/ui/**/*'
+jobs:
+  cypress-run:
+    runs-on: ubuntu-latest
+    container: cypress/browsers:node-20.5.0-chrome-114.0.5735.133-1-ff-114.0.2-edge-114.0.1823.51-1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
Lykhoyda, line 12: there are already 3 version for quite a while, so I would suggest to update it

.github/workflows/cypress-tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+      - uses: actions/cache@v2
```
Lykhoyda, line 14: same here 3 version is available

.github/workflows/cypress-tests.yaml
Tbaut: sure, let's try.

Tbaut: done

packages/ui/src/components/WatchedAccounts.tsx
```diff
@@ -36,7 +36,7 @@ const WatchedAccounts = ({ className }: Props) => {
         })}
       </ListStyled>
       <TitleStyled>Watch new account...</TitleStyled>
-      <AccountSelectionWrapperStyled>
+      <AccountSelectionWrapperStyled data-cy="watched-account-wrapper">
```
Lykhoyda, line 39: I would suggest using Testing library with Cypress, as it allows you to write the selectors that aren't id specific, rather the role. During the test run it assigns the role and you can find an element without artificially creating "data-cy" parameter which is used only by the test. Also, there is a plugin that suggests to you what is the proper selector to choose for certain elements.  (https://chrome.google.com/webstore/detail/testing-playground/hejbmebodbijjdhflfknehhcgaklhano)

Lykhoyda:
> 

Corepack enable is an option that is deactivated by default in Node JS < 16.10. Some node js containers even with higher versions could turn it off as well. So for now there is no guarantee that the node JS version > 18 will have a flag "enable" by default. That's why it's safer to enable the "corepack" functionality.

packages/ui/src/components/WatchedAccounts.tsx
Tbaut: I'm not sure I get it. `data-cy` is actually best practice https://docs.cypress.io/guides/references/best-practices#Selecting-Elements We do want to be very specific, and not rely on elements or classes, or any other id that could be duplicated or used for anything else than testing.

Or did I miss you point?

Tbaut:
re the corepack, my point is that I added the release of yarn in this PR, and that made the Cypress install use yarn v2. I was not able to have `corepack enable` ran, hence added the release which solved my problem.

.github/workflows/cypress-tests.yaml
```diff
@@ -0,0 +1,51 @@
+name: Cypress tests
+on:
+  pull_request:
+    paths:
+      - 'packages/ui/**/*'
+jobs:
+  cypress-run:
+    runs-on: ubuntu-latest
+    container: cypress/browsers:node-20.5.0-chrome-114.0.5735.133-1-ff-114.0.2-edge-114.0.1823.51-1
```
Lykhoyda, line 9: I found in the docs the following container settings (https://github.com/cypress-io/github-action#docker-image):
```
    container:
      image: cypress/browsers:latest
      options: --user 1001
```

We probably would need to use the latest version, I assume specific containers will always have hardcoded version of Cypress

Lykhoyda (changes requested):
I left some comments, I would definitely encourage to use testing library since the beginning and avoid creating the data params in the code

.github/workflows/cypress-tests.yaml
Tbaut: Nice, let's try
```suggestion
    container:
      image: cypress/browsers:latest
      options: --user 1001
```

Tbaut:
Thanks, I answered [in the thread](https://github.com/ChainSafe/Multix/pull/337#discussion_r1317030749). Andrew and I have been working with `data-cy` in the past, since this was deemed best practice (see the link I posted there). Happy to do this differently if this has a clear benefits. Right now adding a `data-cy` to our components is quite annoying to do for the person building the tests, but it makes the test reliable. Can you give an example on how to do differently? I'm not using Chrome and didn't find the extension you mentioned there.

.github/workflows/cypress-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+on:
+  pull_request:
+    paths:
+      - 'packages/ui/**/*'
+jobs:
+  cypress-run:
+    runs-on: ubuntu-latest
+    container:
+      image: cypress/browsers:latest
+      options: --user 1001
```
Tbaut, line 11: getting some permission issues
```suggestion
#      options: --user 1001
```

.github/workflows/cypress-tests.yaml
Tbaut: worked without the options actually.

packages/ui/src/components/WatchedAccounts.tsx
Lykhoyda: @Tbaut 
It's exactly the problem that the library will solve, you don't need to add anything manually even the data-cy ids. It will add the "role" attribute on the run based on the elements (f.e it will assign a role button to the buttons that you can find with the following selector: cy.findByRole('button', 'connect')). Testing library it's actually a gold standard and can be used almost with any frontend framework, It worth the time to learn it once and use it for any project. Here is how to set it up with Cypress https://testing-library.com/docs/cypress-testing-library/intro

The main benefit is that it forces you to write the tests in the way that you interact with visual data, elements, and text and test the app with the way the user will see it. With data- props you can assign it even to the invisible element and still the test will pass as it's present in DOM, but visually it might not appear on the page. More information here:
https://testing-library.com/docs/#the-problem 


Regarding the playground plugin:
It's just a helper tool if you struggle to find a meaningful selector for the element, You can try it here (https://testing-playground.com/), hover over the element and it will show you in the green box how is a good query could look like.

Lykhoyda:
> Thanks, I answered [in the thread](https://github.com/ChainSafe/Multix/pull/337#discussion_r1317030749). Andrew and I have been working with `data-cy` in the past, since this was deemed best practice (see the link I posted there). Happy to do this differently if this has a clear benefits. Right now adding a `data-cy` to our components is quite annoying to do for the person building the tests, but it makes the test reliable. Can you give an example on how to do differently? I'm not using Chrome and didn't find the extension you mentioned there.
> 
> edit: BTW the selector playground from Cyprss does exactly what the extension you mention does, it's backed in Cypress, try it out!

Yeah, I know that it can do it, but it does it awfully. The selectors if it's not a ID will look like autogenerated especially for nested elements

.github/workflows/cypress-tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+    paths:
+      - 'packages/ui/**/*'
+jobs:
+  cypress-run:
+    runs-on: ubuntu-latest
+    container:
+      image: cypress/browsers:latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
Lykhoyda, line 13: does it work with version 4 ? https://github.com/marketplace/actions/checkout

Lykhoyda (changes requested):

.github/workflows/cypress-tests.yaml
Tbaut: it does!

packages/ui/src/components/WatchedAccounts.tsx
Tbaut: Thanks a lot for the detailed, explanation. I think this can help @asnaith a lot, so that you don't have to actually add some `data-cy` all over. @asnaith can you just tell me if you're good with using this, and I'll add it in this PR, and rewrite that first test using it.

PR #338, opened 2023-09-07
Cypress watch address test update

asnaith:
- Changed default test folder from "e2e" to "tests" (more of a gripe with Cypress naming, not all the tests will be e2e and I find it misleading)
- Updated and added to the Watch Account tests


packages/ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -0,0 +1,7 @@
+export const settingsPage = {
+  watchAccountAccountInput: () => cy.findByLabelText('Address'),
+  addButton: () => cy.findByRole('button', { name: /add/i }),
+  multisigIcon: () => cy.get('.identicon'),
+  multisigNameLabel: () => cy.get('.multisigName'),
+  multisigAddressLabel: () => cy.get('.multisigAddress')
```
asnaith, line 6: Hey @Lykhoyda, I would usually just add a data-cy tag to these multisig elements above

CleanShot 2023-09-06 at 22 48 11@2x

I tried using findByRole but these elements don't produce anything to locate by. The only thing I could use without adding a tag is the class name....but that's a Cypress locator, not the testing playground stuff. 

It works fine but I was under the impression that locating elements via class names (with the default cypress locators) was inferior to other locator techniques so didn't want to leave it as is. Do you have any recommendations?

Lykhoyda: @asnaith  It might also happen that there is no meaningful selector and as a last resort, we can use classes to query. However, we can be more specific here, for example, the parent component has a role "List" and we can search for classes inside this container with the "within" method (https://testing-library.com/docs/dom-testing-library/api-within/)

Screenshot 2023-09-07 at 11 48 56

So it will be something like that:
```
cy.findByRole('list').within(() => {
  cy.get('.identicon')
})
```

There are rules about the priority of selectors (https://testing-library.com/docs/queries/about#priority) and if you can't come up with a good selector you can use "data-testid" which is similar to cypress approach with id

```js
// In react app

            <NameStyled
              data-testid={'multisigName'}
              className="multisigName"
            >
              {localName || mainDisplay}
            </NameStyled>

```

```js
cy.findByTestId('multisigName')
```

asnaith: @Lykhoyda ah ok, thank you. 

So just to confirm - when we do have to use a `data-testid` we'd do that instead of `data-cy` so that we can use `getByTestId` from the testing library (rather than having a mix of locators from testing library and cypress)?

packages/ui/src/components/AccountDisplay.tsx
```diff
@@ -71,11 +71,13 @@ const AccountDisplay = ({
...
-        address={encodedAddress}
-        size={iconSize}
-      />
+      <span data-cy="icon-identity">
+        <IdenticonBadge
+          badge={badge}
+          address={encodedAddress}
+          size={iconSize}
+        />
+      </span>
```
asnaith, line 80: @Tbaut I tried everything to avoid adding a span but this is the only thing that would work here (and in the other place where I added one)

Tbaut: This is something we can most probably work around. I'll check it out.

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+import { localHost } from '../../fixtures/landingData'
+import { landingPage } from '../../support/page-objects/landingPage'
+import { settingsPage } from '../../support/page-objects/settingsPage'
+
+describe('Watched Accounts', () => {
+  it('can add an account to the watch list', () => {
+    cy.visit(localHost)
+    landingPage.watchAccountButton().click()
+    settingsPage.accountAddressInput().type(`${addresses.Alice}{enter}`)
+    // autocomplete on the input seems to affect reliability so we wait a bit
```
Tbaut, line 11: This still failed for me :( like the address from Alice was truncated with the first `5`

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+      win.localStorage.setItem('multix.watchedAccount', `["${addresses.AliceSS58}"]`)
+    })
+    cy.visit(localHost)
+    landingPage.watchAccountButton().click()
+    // attempt to add the same account again
+    settingsPage.accountAddressInput().type(`${addresses.Alice}{enter}`)
+    cy.wait(1000)
+    settingsPage.accountNameInput().type('{selectall}{del}Alice')
+    settingsPage.addButton().click()
+    settingsPage.errorLabel().should('be.visible')
```
Tbaut, line 49: We should be specific on this one, and check for the error message. It failed for me for "invalid address", although the check is for something different.

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+      settingsPage.accountIcon().should('not.exist')
+      settingsPage.accountAddressLabel().should('not.exist')
+    })
+    settingsPage.accountContainer().should('have.length', 0)
+  })
+
+  it('can see error when attemping to add same address more than once', () => {
+    // stub existing watched account
+    cy.window().then((win) => {
+      win.localStorage.setItem('multix.watchedAccount', `["${addresses.AliceSS58}"]`)
```
Tbaut, line 40: I think this is dangerous to do as it is messing with localstorage, that shouldn't be messed with. Localstorage isn't something that should be tested. We should manually add an account, and followup by checking we can't add it twice.

I think it's extremely unlikely that someone pastes the public key (I think the name here is misleading BTW, ss58 is something different). Maybe we should have Alice with some other ss58 encoding?

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+      settingsPage.accountIcon().should('be.visible')
+      settingsPage.accountAddressLabel().should('be.visible')
+      settingsPage.accountNameLabel().should('be.visible')
+      settingsPage.accountDeleteButton().should('be.visible')
+    })
+  })
+
+  it('can remove an account from the watch list', () => {
+    cy.window().then((win) => {
+      win.localStorage.setItem('multix.watchedAccount', `["${addresses.AliceSS58}"]`)
```
Tbaut, line 25: Same here, I think tests should be independant from the implementation details. In this case, adding an account is super quick and we should rather use the UI for this.

Tbaut (commented):
I had a look, and things fails for me unfortunately due to this drop-down somehow deleting some text :shrug: 

I'll go around and fix those `cy-data` too, because adding span is dangerous with regard to the rendering of the app. I think we can avoid most of them with some tiny changes.

packages/ui/cypress/fixtures/accounts.ts
```diff
@@ -1,3 +1,4 @@
 export const addresses = {
-  Alice: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'
+  Alice: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY',
+  AliceSS58: '0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d'
```
Tbaut, line 3: I'd suggest something like this.
```suggestion
  AlicePubKey: '0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d',
  AliceKusama:'HNZata7iMYWmk5RvZRTiAsSDhV8366zq2YGb3tLH5Upf74F'
```

Tbaut:
It doesn't make any sense to me.. but adding video recording in case of failed test to try to understand what was up with those tests in prod that I couldn't reproduce locally.. it now passes :roll_eyes:

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
asnaith: @Tbaut I was going to....but then remembered back to when I did that on another project and we said we didn't want to check the actual string 😄. I will take a look and add this in

packages/ui/cypress/fixtures/accounts.ts
asnaith: I'll remove it, not needed if setting everything up in the UI

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
asnaith: It seems to be the speed of cypress typing and when the autocomplete / dropdown appears that makes it wonky. I can slightly slow down the speed of typing and it helps. Best I can do at the moment as can't find any event, attribute or request to spy on.

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
asnaith: Updated and improved now, check the string and also that the add button is disabled

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
asnaith: Updated to do everything within the UI now

packages/ui/cypress/tests/watched-acccounts/watched-accounts.cy.ts
asnaith: Updated here too

packages/ui/cypress/fixtures/accounts.ts
asnaith: Removed :)

packages/ui/cypress.config.ts
```diff
@@ -2,6 +2,7 @@ import { defineConfig } from 'cypress'
 
 export default defineConfig({
   e2e: {
+    specPattern: 'cypress/tests/**/*.cy.{js,jsx,ts,tsx}',
```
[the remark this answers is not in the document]
asnaith: Good point, updated. We only need .ts

asnaith:
> It doesn't make any sense to me.. but adding video recording in case of failed test to try to understand what was up with those tests in prod that I couldn't reproduce locally.. it now passes 🙄 In any case, we got rid of those wait 🚀

I was still having issues running locally on macOS once the wait was removed and couldn't find anything to spy on to make it more reliable. I was able to add a small 20ms delay on the keypress which works but really that's just a less explicit wait

Tbaut (commented):
oops had those comments not submitted. I'll address them and merge I guess.

Tbaut:
Moved the redundant code into a function. Added a settings page to visit.

Tbaut (approved):
Looks great, thank you so much!

PR #340, opened 2023-09-07
Add WC id to clouflare deployment

Tbaut:
I forgot to push the WC id to cloudflare, so it's not connecting to it in prod.


Tbaut:
merging because in prod.

Issue #342, opened 2023-09-08
Add Cypress Extension connection (and account sharing) 


Tbaut:
This is the first step of a long journey.
We want to allow Cypress users to get `cy` commands to check and verify whether there has been wallet connection request, and answer with the accounts they want.

FWIW I start here, and will extract this into another repo dedicated to the plugin later on, when the basic features are in place.

PR #344, opened 2023-09-11
Programmatic login with an extension

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/342
Adds some commands to Cypress to:
- initialize the extension and set some accounts
- view the authorization requests from the Dapp and check the origin name (that's the only info transited from the Dapp)
- accept the connection from a give request, and tell what accounts should be shared with the Dapp


packages/ui/cypress/support/commands.ts
```diff
@@ -39,3 +41,47 @@ import '@testing-library/cypress/add-commands'
...
+const extension = new Extension()
+
+Cypress.Commands.add('initExtension', (accounts: InjectedAccount[]) => {
+  cy.log('Initializing extension')
+  cy.window().then((win) => {
+    Object.defineProperty(win, 'injectedWeb3', {
+      get: () => extension.init(accounts)
+    })
+  })
+})
```
asnaith, line 54: 🙌

packages/ui/cypress/e2e/login.cy.ts
```diff
@@ -0,0 +1,24 @@
+import { injectedAccounts } from '../fixtures/injectedAccounts'
+import { localHost } from '../fixtures/landingData'
+
+describe('login-accounts', () => {
```
asnaith, line 4:
```suggestion
describe('Connect Account', () => {
```
This is just what appears in Cypress / any reporting. I usually write it like this

packages/ui/cypress/e2e/login.cy.ts
```diff
@@ -0,0 +1,24 @@
...
+  beforeEach(() => {
+    cy.visit(localHost)
+  })
+
+  it('Connects with Alice', () => {
+    cy.initExtension(injectedAccounts)
+    cy.get('[data-cy="button-menu-connect"]').click()
```
asnaith, line 11: I'd move element `get` like this to a page object but we can tidy up later

packages/ui/cypress/e2e/login.cy.ts
```diff
@@ -0,0 +1,24 @@
...
+      cy.wrap(requests.length).should('eq', 1)
+      // this request should be from the application Multix
+      cy.wrap(requests[0].origin).should('eq', 'Multix')
+      // let's allow it for Alice
+      cy.enableAuth(requests[0].id, [Object.values(injectedAccounts)[0].address])
+      // the ui should then move on to connecting to the rpcs
+      cy.get('[data-cy="loader-rpc-connection"]').should('contain', 'Connecting to the node at')
+    })
+  })
+})
```
asnaith, line 24: This is awesome as an explicit connect test and we can later wrap all the steps into a helper function we call as a pre-cursor to most tests! 🙌

asnaith:
This is fantastic, a huge unlock! 

I think it's best to leave the full review of the code to @Lykhoyda as there are a few things here that I'm not sure about but functionally looks great and works really well.

packages/ui/cypress/e2e/login.cy.ts
Tbaut: sure

packages/ui/cypress/e2e/login.cy.ts
Tbaut: absolutely

asnaith (approved):
> This is fantastic, a huge unlock!
> 
> I think it's best to leave the full review of the code to @Lykhoyda as there are a few things here that I'm not sure about but functionally looks great and works really well.

Hmmm, actually I changed my mind. I'm ok with approving even though I don't understand every single line yet because it's safe to do so - not the final version and it's test code, we already know it will be iterated on further. Let's go 🚀

packages/ui/src/contexts/AccountsContext.tsx
```diff
@@ -39,6 +39,8 @@ const AccountContextProvider = ({ children }: AccountContextProps) => {
   const [extensions, setExtensions] = useState<InjectedExtension[] | undefined>()
   const [timeoutElapsed, setTimoutElapsed] = useState(false)
   const { chainInfo } = useApi()
+  console.log('isAccountLoading', isAccountLoading)
+  console.log('isExtensionError', isExtensionError)
```
Tbaut, line 43:
```suggestion
```

PR #346, opened 2023-09-12
feat: update header view for solo multisig

### LINKED ISSUE #304: Use the main header for multisigs and pure proxies
User feedback:
> the home layout is different for one with and without pure proxy and that's inconsistent. like the new transaction button position is different 

We use a large header currently only when there's a proxy. We should also use it for when there's a solo multisig.

Design: https://www.figma.com/file/4dQC9kxf5uxCrAjpkzAK0D/Multix?node-id=1005%3A17277&mode=dev
image

Lykhoyda:
Closes #304 

- Added header view for multisig
- Renamed component from PureHeaderView -> HeaderView


asnaith:
The new header looks good :) 

It kind of makes the pre-existing section seem a little bit strange though if we were to merge as it is (two new transaction buttons etc). 

I'm not sure what the solution should be. If we replace it with "controlled by" (same as we show when the pure is selected) we'd still be showing the multisig identicon twice. 

Maybe @sweetpea22 has a suggestion? 🤞

CleanShot 2023-09-12 at 16 40 48@2x

Tbaut:
I think this PR isn't ready to be reviewed since the original issue has the design and I think it's clear and we should 100% not have duplication of info.
image

sweetpea22:
hiya, the pic up there is an older design. i think we designed we wanted the balance not to be a block since it looked a bit weird. here's an updated version from what we discussed, should we go with this one instead?
Screen Shot 2023-09-13 at 7 18 37 PM

asnaith:
I think just a string change is remaining now (to say "Account Details")

CleanShot 2023-09-15 at 10 40 40@2x

Tbaut:
Please check the design from Cindy above and my comment. There is no need to show the multisig again, it's already in the header.

asnaith (approved):
Latest changes look great :)

asnaith (dismissed):
Oops sorry, I just realized the new transaction button is showing on the header for watched accounts

packages/ui/src/pages/Home/HeaderView.tsx
```diff
@@ -31,15 +32,19 @@ const PureProxyHeaderView = () => {
...
+          <BalanceStyled>
+            <BalanceHeaderStyled>Balance</BalanceHeaderStyled>
+            <BalanceAmountStyled>
+              <Balance address={selectedAddress} />
+            </BalanceAmountStyled>
+          </BalanceStyled>
+        </BalanceStyledWrapper>
         <BoxStyled>
-          <MultisigActionMenu withSendButton={!selectedIsWatched} />
+          <MultisigActionMenu />
```
Tbaut: This should stay
```suggestion
          <MultisigActionMenu withSendButton={!selectedIsWatched}/>
```

packages/ui/src/pages/Home/MultisigActionMenu.tsx
```diff
@@ -45,30 +49,48 @@ const MultisigActionMenu = ({ withSendButton = true }: MultisigActionMenuProps)
...
 
+    // If we are NOT rendering "new transaction button", the functionality will appear in the list
+    if (!withNewTransactionButton) {
+      opts.unshift({
+        text: 'Send transaction',
+        icon: <HiOutlinePaperAirplane size={20} />,
+        onClick: () => setIsSendModalOpen(true)
+      })
+    }
+
```
Tbaut: This is not needed (for now at least) 
```suggestion
```

packages/ui/src/pages/Home/MultisigActionMenu.tsx
```diff
@@ -45,30 +49,48 @@ const MultisigActionMenu = ({ withSendButton = true }: MultisigActionMenuProps)
...
+        menuButtonBorder={menuButtonBorder}
+        options={options}
+      />
     </>
   )
 }
 
+MultisigActionMenu.defaultProps = {
+  withNewTransactionButton: true
+}
```
Tbaut, line 95: We've never used this in the past. Please keep the `{ withSendButton = true }` in the props, which shows right away what the default is, and we can remove this.
```suggestion
```

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -4,52 +4,55 @@ import MultisigActionMenu from './MultisigActionMenu'
...
-            {selectedHasProxy
-              ? 'Controlled by'
-              : renderMultisigHeading(selectedMultiProxy.multisigs.length > 1)}
-          </h3>
-        )}
-        {!selectedHasProxy && (
+        {selectedMultiProxy && <h3>{selectedHasProxy ? 'Controlled by' : 'Account Details'}</h3>}
+        {selectedHasProxy && (
           // only show the buttons if we have a solo multisig
           // and only show the send button if we are part of the multisig
```
Tbaut, line 19: I don't think this comment makes sense any more.
```suggestion
```

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -4,52 +4,55 @@ import MultisigActionMenu from './MultisigActionMenu'
...
           <BoxStyled>
-            <MultisigActionMenu withSendButton={!selectedIsWatched} />
+            <MultisigActionMenu
+              menuButtonBorder={'none'}
+              withNewTransactionButton={false}
+            />
```
Tbaut, line 24: Unless I'm mistaking, we should actually never have this action buttons any more. At least not now. Maybe we should comment it out.

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -4,52 +4,55 @@ import MultisigActionMenu from './MultisigActionMenu'
...
-              <MultisigWrapperStyled key={multisig.address}>
-                <AccountDisplayWrapperStyled>
-                  <AccountDisplay
-                    address={multisig.address || ''}
-                    badge={AccountBadge.MULTI}
-                    withBalance={false}
-                  />
-                </AccountDisplayWrapperStyled>
+              <MultisigWrapperStyled
+                isPureMultisig={selectedHasProxy}
```
Tbaut, line 33: can we keep the same naming?
```suggestion
                selectedHasProxy={selectedHasProxy}
```

Tbaut (changes requested):

packages/ui/src/pages/Home/MultisigActionMenu.tsx
```diff
@@ -45,30 +49,49 @@ const MultisigActionMenu = ({ withSendButton = true }: MultisigActionMenuProps)
...
+    // If we are NOT rendering "new transaction button" and is it's NOT a Watched account,
+    // the "Send transaction" item will appear in the list menu
+    if (!withNewTransactionButton && !selectedIsWatched) {
+      opts.unshift({
+        text: 'Send transaction',
+        icon: <HiOutlinePaperAirplane size={20} />,
+        onClick: () => setIsSendModalOpen(true)
+      })
+    }
+
```
Tbaut, line 61: I don't think this is needed, since there's nothing in the 3 dots buttons that should change.
```suggestion
```

packages/ui/src/pages/Home/MultisigActionMenu.tsx
Lykhoyda: Yeah, it's required only when we have a separate button for each multisig. I will comment this line

packages/ui/src/pages/Home/MultisigView.tsx
Lykhoyda: Actually, I am considering having a descriptive name for the conditions we put for multisig or pure multisig as the name selectedHasProxy requires knowledge of the project to understand from where the word is coming.

Every time I do the condition I should mentally remember:
it has a proxy => which means it's pure multisig => because only pure multisig can have proxy

Instead, we could return "type" from Context

export interface IMultisigContext {
   multisigType: "pure" | "solo"
}

For current PR I will use selectedHasProxy

packages/ui/src/pages/Home/MultisigActionMenu.tsx
Lykhoyda: Commented for the individual multisig button issue

packages/ui/src/pages/Home/MultisigView.tsx
Tbaut: I called the overall object MultiProxy, because it's a tree with "pure > multisigs []" where the pure could be empty. We could for sure add some more derived state, but this may make this tree a little bloated maybe? So far we mostly needed to know whether it's a pure (with several multisigs) or if it's a single Multisig, and this is in the `selectedHasPure`. Similarly we have `selectedIsWatched`. The main thing I see is that we'd need eventually a `selectedDisplayAddress`, which is some code that's duplicated with `selectedIsWatched ? selected.pure : selected.multisigs[0].address`. 

But yes this should 100% be for another PR.

Tbaut (commented):
awesome. Just some minor changes that I'll commit right away to merge

Tbaut (approved):
Checked as many edge cases as I could, it's looking great, thank you so much!

PR #348, opened 2023-09-14
Make account connection test robust

Tbaut:
The test wasn't robust because the code wasn't either. It had a trick to re-trigger the connection request, which made the test flaky. It should now be cleaner, and stable


asnaith (approved):
Excellent! Thank you for the stabilization fixes 🚀

PR #349, opened 2023-09-18
Add moonbeam/river to the dropdown

Tbaut:
The networks have been tested extensively in #334 this now makes them officially supported because visible in the dropdown.


asnaith (approved):
Good to go. I created a multisig successfully on Moonbeam

PR #355, opened 2023-09-19
Don't show errors when submitting extrinsics

### LINKED ISSUE #350: Minimum balance warning appears during creation
**Description**
If an account already has the minimum required balance to create a multisig when it takes this during the creation phase, before it's complete, the minimum amount warning banner is displayed whilst it is completing

**Prerequisites**

Use an account that only has the minimum amount required as balance 

**Steps To Reproduce**

1. Go to Multix at 'https://multix.chainsafe.io/'
2. Go through the multisig creation 
3. Observe that before it's complete the warning about minimum balance will appear 

**Expected Behaviour**
Once the multisig is being created we should not do any minimum balance checks

Tbaut:
closes #350


asnaith (approved):
Looking good :)

PR #356, opened 2023-09-21
Make better errors

### LINKED ISSUE #351: Show balance warning on approval modal
**Description**
There is no balance warning on the approval modal when a user doesn't have enough funds to actually approve it

**Steps To Reproduce**

1. Try and approve a transaction with an account that has no funds, observe the user can make the approval (but will then see a toast failure because of not enough funds)

Tbaut:
closes #351 
And on top of this it
- Shows reserved amount on the summary for multisig creation and multisig changes
- Says the creation takes ~1min (it's more 30s on relay chains, but 10min
- Set errors on top of the modal overlay (see screenshot)

image


asnaith (approved):
Tested the original scenario and the new additions, all working as intended :)

PR #360, opened 2023-09-21
Bump UI dependancies

Tbaut:
- bump deps
- remove `@testing-library/cypress`


packages/ui/src/index.tsx
```diff
@@ -11,7 +11,9 @@ const container = document.getElementById('root')
...
 
 root.render(
-  <StrictMode>
-    <RouterProvider router={router} />
-  </StrictMode>
+  (
+    <StrictMode>
+      <RouterProvider router={router} />
+    </StrictMode>
+  ) as any
```
Tbaut, line 18: @Lykhoyda if you've an idea how to remove this, that'd be appreciated.. I sounds like the `render` isn't happy with the return of the StrictMode.. although this dep hasn't actually changed.

Lykhoyda: @Tbaut  StictMode should stay in place, it helps with the development. Can you please provide the error output that you are facing?

Tbaut: Right, I never meant to change the strict mode. The error is
> Argument of type 'Element' is not assignable to parameter of type 'ReactNode'.

This error needs context though, you can check it out by just removing the `as any`. I'm on the fence between the `any` or downgrading `react-router-dom`.. having a hard time

Tbaut: spent too much time on this.. I really don't know where it originates from. typescript and react-router-dom rollback doesn't help. I'd just merge as is.

Tbaut: Thanks a lot for checking it out. Your fix worked, and pointed me in the direction that we actually don't even need the `@types/react`. Removing it solved the issue too :man_shrugging:

PR #361, opened 2023-09-22
Add accordion to settings page

### LINKED ISSUE #336: Rework Settings view to use accordion
The Settings should be split between categories, with accordions
image

The design is here: https://www.figma.com/file/4dQC9kxf5uxCrAjpkzAK0D/Multix?node-id=1076%3A23454&mode=dev

Also, we should be able to jump to a specific section direction with a link (using anchor `settings#watched-acccounts` ?). E.g we have buttons bringing users directly to "Watched accounts". When they click on such a button, they should navigate to the settings and this part should be expanded.

Lykhoyda:
Closes #336 
- Update UI of the Settings page (Watched account and Wallet connect)
- Watched accounts moved from components to the Settings page folder (as used now only in one place)
- Feature to expand accordion using hashtags #watched-acccounts and #wallet-connect

! 
There is no Wallet connect design yet


packages/ui/src/pages/Settings/WatchedAccounts.tsx
```diff
@@ -0,0 +1,135 @@
...
+            md={8}
+          >
+            <PaperStyled>
+              {watchedAddresses.map((address, index) => {
+                return (
+                  <AccountItemWrapperStyled
+                    key={address}
+                    data-cy="container-account-details"
+                  >
+                    <AccountCountStyled>{index + 1}</AccountCountStyled>
```
Tbaut, line 36: Sorry that I didn't see this before in the design, but this is not needed, as it takes space and doesn't add value.
```suggestion
                    <AccountCountStyled>{index + 1}</AccountCountStyled>
```

Tbaut (commented):
This looks great!! I've a couple small changes that I'll push, like for the WalletConnect icon.

Also the links to setting > watched account need to be modified in the home page (and maybe some other places?) with the `#panel-watched-accounts`. This should make the Cypress test pass, or get us closer to.

packages/ui/src/pages/Settings/Settings.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { WalletConnectActiveSessions } from '../../components/WalletConnect/WalletConnectActiveSessions'
+import { HiOutlineChevronDown as ExpandMoreIcon, HiOutlineEye } from 'react-icons/hi2'
+import { theme } from '../../styles/theme'
+import { SyntheticEvent, useEffect, useState } from 'react'
+import { useLocation } from 'react-router-dom'
+
+const Settings = () => {
+  const location = useLocation()
+  const [expanded, setExpanded] = useState<string | false>(false)
+  const handleChange = (panel: string) => (event: SyntheticEvent | null, isExpanded: boolean) => {
```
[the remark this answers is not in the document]
Lykhoyda: Wrapped in callback. I'm not sure about simplifying the methods as we have it as a callback from Accordion

```js
onChange?: (event: React.SyntheticEvent, expanded: boolean) => void;
```
it's also an example from the documentation, so if something is not clear there are more details by the link https://mui.com/material-ui/react-accordion/#system-ControlledAccordions.tsx

packages/ui/src/pages/Settings/Settings.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { WalletConnectSession } from '../../components/WalletConnect/WalletConnectSession'
+import { WalletConnectActiveSessions } from '../../components/WalletConnect/WalletConnectActiveSessions'
+import { HiOutlineChevronDown as ExpandMoreIcon, HiOutlineEye } from 'react-icons/hi2'
+import { theme } from '../../styles/theme'
+import { SyntheticEvent, useEffect, useState } from 'react'
+import { useLocation } from 'react-router-dom'
+
+const Settings = () => {
+  const location = useLocation()
+  const [expanded, setExpanded] = useState<string | false>(false)
```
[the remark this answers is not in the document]
Lykhoyda: I think we could be more precise as we can have the name of the accordion, which corresponds to a true or a collapsed boolean state.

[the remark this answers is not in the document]
Lykhoyda: Created a type of strings

Lykhoyda:
> This looks great!! I've a couple small changes that I'll push, like for the WalletConnect icon. I'll do it later today.
> 
> The links to setting > watched account need to be modified in the home page (and maybe some other places?) with the `#panel-watched-accounts`. This should make the Cypress test pass, or get us closer to.

We need to add click on accordion for cypress test as before the inputs appeared right away.

Lykhoyda:
@asnaith @Tbaut please take a look again. I fixed the comments, let me know if i missed something

packages/ui/src/pages/Settings/WatchedAccounts.tsx
```diff
@@ -4,36 +4,34 @@ import { useWatchedAddresses } from '../../contexts/WatchedAddressesContext'
...
+import { useCallback } from 'react'
 
-interface Props {
-  className?: string
-}
-
-const WatchedAccounts = ({ className }: Props) => {
+const WatchedAccounts = () => {
   const { watchedAddresses, removeWatchedAccount, addWatchedAccount } = useWatchedAddresses()
+  const hasWatchedAddresses = useCallback(() => watchedAddresses.length > 0, [watchedAddresses])
```
Tbaut, line 11: This should be a `useMemo`

packages/ui/src/pages/Settings/WatchedAccounts.tsx
```diff
@@ -4,36 +4,34 @@ import { useWatchedAddresses } from '../../contexts/WatchedAddressesContext'
...
 
   return (
     <>
-      <WatchAccountsHeaderStyled>Currently watched accounts:</WatchAccountsHeaderStyled>
+      {hasWatchedAddresses() && (
```
Tbaut, line 15: then `hasWatchedAddresses` without needing to call it.

packages/ui/src/pages/Settings/WatchedAccounts.tsx
```diff
@@ -4,36 +4,34 @@ import { useWatchedAddresses } from '../../contexts/WatchedAddressesContext'
...
+        <WatchAccountsHeaderStyled>Currently watched accounts:</WatchAccountsHeaderStyled>
+      )}
       <Grid
-        className={className}
         container
         spacing={2}
       >
-        {watchedAddresses.length > 0 && (
+        {hasWatchedAddresses() && (
```
Tbaut, line 22: same here.

packages/ui/src/components/WalletConnect/WalletConnectSession.tsx
```diff
@@ -1,6 +1,6 @@
 import { Alert, CircularProgress, Grid, styled } from '@mui/material'
 import { Button, TextFieldStyled } from '../library'
-import { useCallback, useMemo, useState } from 'react'
+import React, { useCallback, useMemo, useState } from 'react'
```
Tbaut, line 3: No need for this.

Tbaut (changes requested):

asnaith:
> @asnaith @Tbaut please take a look again. I fixed the comments, let me know if i missed something

@Lykhoyda Those updates for the labels look good 👍

> We need to add click on accordion for cypress test as before the inputs appeared right away.

If you can update the link for the button below to use settings#watched-acccounts then it's going to be expanded when arriving on that screen and will be better for users and also no need to update the test

CleanShot 2023-09-22 at 17 27 29@2x

packages/ui/src/pages/Settings/Settings.tsx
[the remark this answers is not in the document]
Tbaut: Changed it in my PR. This was definitely needlessly complex :man_shrugging:

Tbaut:
I pushed some change to
- change a pit the logic to force open only the watched-account section if needed.
- optimize the WC SVG that took a while to load
- make the tests pass. Please check it out. We won't merge any PR that doesn't pass the test, unless it's due to flakiness (not the case here).

There are still a couple comments above that weren't taken into account yet, but it should be quick.

Tbaut (approved):
awesome, this looks a million time better. Thank you!

Issue #264, opened 2023-07-18
Rework the network dropdown


Tbaut:
With the growing amount of parachains being supported, the simple dropdown for networks will soon not suffice.
I propose to switch from a simple select, to something more advanced. for instance what Polkassembly is proposing:
![image](https://github.com/ChainSafe/Multix/assets/33178835/da63b2ce-9216-4e37-971a-1874215b5df2)

Or subscan with a button and Modal:
![image](https://github.com/ChainSafe/Multix/assets/33178835/d9f82789-fe76-4c5d-8592-6aa77e4a3541)

Tbaut:
Also adding to this, there's this annoyance in the current dropdown, noticed by a user
> the clickable area is only the icon and text, not including the margin

https://github.com/ChainSafe/Multix/assets/33178835/2d8e03e0-8847-421e-bdc0-90e423f836b6

asnaith:
Let's include improving this too? 

<img width="224" alt="CleanShot 2023-08-18 at 09 44 45@2x" src="https://github.com/ChainSafe/Multix/assets/3982665/0cae48ac-3290-488e-9f47-a42b2caa9815">

<img width="195" alt="CleanShot 2023-08-18 at 09 45 11@2x" src="https://github.com/ChainSafe/Multix/assets/3982665/aeab52ad-5429-4dea-a97f-aac7b0ac061a">

Tbaut:
Sure

Lykhoyda:
Make sure that the menu fits for the range of screens 900 < 1078 and the menu items are aligned in one line

<img width="469" alt="Screenshot 2023-09-26 at 12 49 58" src="https://github.com/ChainSafe/Multix/assets/8141917/78792e88-7886-415b-9096-4ec5d3160eda">

Tbaut:
The selector can be very small, just the icon even such as https://www.figma.com/file/4dQC9kxf5uxCrAjpkzAK0D/Multix?node-id=210%3A7442&mode=dev so that we leave space for the rest.
![image](https://github.com/ChainSafe/Multix/assets/33178835/dcf1ab72-9e78-401a-9fe7-2df633ceab7b)

PR #366, opened 2023-09-26
Update Header styles

### LINKED ISSUE #260: Menu bar style colors change
We want to come close to this
image

Figma: https://www.figma.com/file/4dQC9kxf5uxCrAjpkzAK0D/Multix?node-id=839%3A27689&mode=dev

This is for:
- changing the logo. Use this: 
MULTIX_LOGO_FULL_BLUE_1200px
- changing the font weigth of the menu
- changing the background for the right gray

Lykhoyda:
Closes #260 

- Updated navbar background
- logo update
- fix navbar elements to jump to the next line
- fix colors, hover
- renamed Header component

updates:
- Fixed the gray background color of the page
- Logo color
- Watch the account navigation link for ConnectOrWatch component


Tbaut (changes requested):
- The background of the main app is `#FAFCFD` afaics from the design. Only the menu (and some components are `#fff` or `white` also on this, it's quite rare in design to use pure white/black. Let me catch up on that with Cindy. In the mean time, we can use white/black cf https://graphicdesign.stackexchange.com/questions/25356/why-not-use-pure-black-000-and-pure-white-fff

- Some elements like a loader are invisible now, there's a loader there usually. I believe it is https://github.com/ChainSafe/Multix/blob/ec101c7dfebe29a2c30e78b84f11ee094feac17e/packages/ui/src/components/select/MultiProxySelection.tsx#L152
image


- Having a scrollbar and the dropdown for Networks cut is worse than having the menu on 2 lines IMHO. We should implement this only after https://github.com/ChainSafe/Multix/issues/264

image

- Let's try not to use `black` or `white` in the code directly, and rather use the theme.palette colors so that we can change them quickly.

Lykhoyda:
> * The background of the main app is `#FAFCFD` afaics from the design. Only the menu (and some components are `#fff` or `white` also on this, it's quite rare in design to use pure white/black. Let me catch up on that with Cindy. In the mean time, we can use white/black cf https://graphicdesign.stackexchange.com/questions/25356/why-not-use-pure-black-000-and-pure-white-fff
> * Some elements like a loader are invisible now, there's a loader there usually. I believe it is https://github.com/ChainSafe/Multix/blob/ec101c7dfebe29a2c30e78b84f11ee094feac17e/packages/ui/src/components/select/MultiProxySelection.tsx#L152
>   
>   image
> * Having a horizontal scrollbar and the dropdown for Networks cut is worse than having the menu on 2 lines IMHO. We should implement this only after [Rework the network dropdown #264](https://github.com/ChainSafe/Multix/issues/264)
>   image
> * Let's try not to use `black` or `white` in the code directly, and rather use the theme.palette colors so that we can change them quickly.

The background color is probably not updated everywhere, please provide the link next time if there is a original version of the design.
2) I agree that is not ideal, but as I said it shouldn't affect the users as screen size is not common. Let's have 2 lines if it's "better" 😃

Tbaut:
> The background color is probably not updated everywhere, please provide the link next time

The link is in the original issue, pointing to a figma node that has been deleted in the mean time... Feel free to ask then. However the figma with this view is still there, and the screenshot in the original issue is also quite clear. On top of this there's a mention of "updating the background to the right gray". Doing my best to have well scoped issues, but at some point you gotta help too :)

Tbaut (approved):
Looking great thanks a lot!

PR #367, opened 2023-09-26
Cypress can now sign programmatically

### LINKED ISSUE #357: [Cypress] Allow to sign transactions
Following what has been done for the login cypress polkadot plugin should allow to
- see a list of submitted transaction
- ideally check for the payload
- sign with an account that the extension/plugin was initiated with

Tbaut:
closes #357

- can sign/reject Txs
- can see the transactions queue and analyze the payload

todo:
- tests will fail if a tx is the same as one that's in the queue so we need to clean pending TXs first if we want to see the whole flow. I'll do this in another PR, when approving/rejecting pending TXs.


packages/ui/cypress/tests/transactions.cy.ts
```diff
@@ -0,0 +1,66 @@
...
+    })
+  })
+
+  it('Abort a tx with Alice', () => {
+    multisigPage.newTransactionButton().click()
+    sendTxModal.sendTxTitle().should('be.visible')
+    sendTxModal.fieldTo().click().type(`${AliceAddress}{enter}`)
+    sendTxModal.fieldAmount().click().type('0.001')
+    sendTxModal.buttonSend().should('be.enabled').click()
+    cy.getTxRequests().then((req) => {
```
Lykhoyda, line 36: I see that you created a class Extension and all the interactions will be mocked as we don't want to deal with the real Polkadot plugin. Is it a correct understanding?

packages/ui/src/components/Toasts/ToastContent.tsx
```diff
@@ -20,14 +20,26 @@ const ToastContent = ({ className, toast }: ToastContentProps) => {
...
+    <div
+      className={className}
+      data-cy="notification-wrapper"
+    >
       <div className="iconContainer">
         {type === 'success' ? (
-          <HiOutlineCheckCircle size={25} />
+          <HiOutlineCheckCircle
+            data-cy="notification-icon-success"
+            size={25}
```
Lykhoyda, line 31: It's a small thing for sure but the sizes are normally even, so 24 is probably a good size

Lykhoyda (changes requested):
Overall looks good considering the accept transaction will be handled in the next PR. I requested some changes for code duplication.

packages/ui/cypress/tests/transactions.cy.ts
```diff
@@ -0,0 +1,66 @@
...
+      const requests = Object.values(authRequests)
+      // we should have 1 connection request to the extension
+      cy.wrap(requests.length).should('eq', 1)
+      // this request should be from the application Multix
+      cy.wrap(requests[0].origin).should('eq', 'Multix')
+      // let's allow it for Alice
+      cy.enableAuth(requests[0].id, [AliceAddress])
+      // the ui should then move on to connecting to the rpcs
+      topMenuItems.multiproxySelector().should('be.visible')
+    })
```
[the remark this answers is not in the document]
Tbaut: Yes, this is the subject of an issue I opened here, to be fixed separately as not super trivial https://github.com/ChainSafe/Multix/issues/371

packages/ui/cypress/tests/transactions.cy.ts
Tbaut: Yes, that's the idea behind the "Cypress plugin" we're building, because you can't install plugins in Cypress. This logic will end up in a separate repo, published to NPM, and imported here, so that others can use it.

packages/ui/src/components/Toasts/ToastContent.tsx
Tbaut: unrelated, but small enough that we can squeeze this in, pushed the change :+1:

packages/ui/cypress/tests/transactions.cy.ts
```diff
@@ -0,0 +1,66 @@
...
+  })
+
+  it('Abort a tx with Alice', () => {
+    multisigPage.newTransactionButton().click()
+    sendTxModal.sendTxTitle().should('be.visible')
+    sendTxModal.fieldTo().click().type(`${AliceAddress}{enter}`)
+    sendTxModal.fieldAmount().click().type('0.001')
+    sendTxModal.buttonSend().should('be.enabled').click()
```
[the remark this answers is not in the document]
Tbaut: Good point, extracted this into `fillAndSubmitTransactionForm` because we click on the send button too.

Tbaut:
The accept transaction is definitely working, but under some specific conditions. I'm on the fence between:
- fix it in the issue I mentioned (https://github.com/ChainSafe/Multix/issues/370)
or
- actually do it right from the get go, and go with a full setup including forking the network with [Chopsticks](https://github.com/AcalaNetwork/chopsticks), running our an indexer against it, start with a blank page every single test, test the multisig creation flow, and signing a TX. This would probably be ideal, but non trivial, and the sequence in which tests will be run will matter.

Lykhoyda (approved):

PR #376, opened 2023-09-27
Balance conversion for balance types

### LINKED ISSUE #375: Make sure all the `Balanceof<>` are denomitated in Token
Following the non-merge of https://github.com/ChainSafe/Multix/pull/312
We should check the type of the field, and if it's `BalanceOf<` or anything with a token, we should show an input with token, and multiply it in the extrinsic by the token decimal.

Tbaut:
closes #375 

This adds the token at the end of the field for manual extrinsics in case the field is expecting a balance. It will then convert the value in plank with the right amount of 0s.

To test:
- Craft a manual extrinsic that needs a value, for instance `balances.transfer` or `bounties.proposeBounty`
- check that the token is visible
- check that when the tx is done, the right amount in plank is submitted. Note that for balances.transfer, the overview converts it back to human readable values, like in my screenshot

Submitting
image

Verifying
image


packages/ui/src/components/EasySetup/ManualExtrinsic.tsx
```diff
@@ -13,19 +14,22 @@ import { ISubmittableResult } from '@polkadot/types/types'
 import React, { ChangeEvent, useCallback, useEffect, useMemo, useState } from 'react'
 import { useApi } from '../../contexts/ApiContext'
 import paramConversion from '../../utils/paramConversion'
-import { getTypeDef } from '@polkadot/types/create'
+// import { getTypeDef } from '@polkadot/types/create'
```
Lykhoyda, line 17: to remove

Tbaut: I was on the fence with this one, because it's super hard to find :smile: my IDE didn't add it automatically, and so because it goes with the other comments (the console.logs that are pretty useful to quickly un-comment rather than find and re-create) I left it. 

Let's remove it, and hope it's not too hard to find again.
```suggestion
```

packages/ui/src/components/EasySetup/ManualExtrinsic.tsx
```diff
@@ -137,11 +90,84 @@ const ManualExtrinsic = ({
...
+          converted = converted.map((single: any) =>
+            isNumType(type)
+              ? single.indexOf('.') >= 0
+                ? Number.parseFloat(single)
+                : Number.parseInt(single)
+              : single
+          )
+          return [...previousValue, converted]
+        }
+
```
Lykhoyda, line 133: It's actually super complex to read and understand. It would be nice to refactor the lines of code by human-readable actions As an example of refactoring such code is easier to understand when the code of each data manipulation is separated by the methods. This way the method gives you understanding of what is suppose to happen without digging into details

```js

const transformParams = useCallback(
  (paramFields, inputParams, opts = { emptyAsNull: true }) => {
    const params = paramFields.map((field, ind) => ({
      ...field,
      value: processParamValue(inputParams[ind], opts.emptyAsNull),
    }));

    return params.reduce((previousValue, { type = 'string', value, typeName }) => {
      if (isEmptyValue(value)) return opts.emptyAsNull ? [...previousValue, null] : previousValue;

      let converted = value;

      if (isVecType(type)) {
        converted = convertVector(value, type);
        return [...previousValue, converted];
      }

      if (isTypeBalance(typeName) && shouldConvertToBalance(chainInfo, converted)) {
        return handleBalanceConversion(previousValue, converted);
      }

      if (isNumType(type)) {
        converted = convertToNumber(converted);
      }

      return [...previousValue, converted];
    }, []);
  },
  [chainInfo, onSetErrorMessage]
);

function processParamValue(inputParam, emptyAsNull) {
  if (typeof inputParam === 'object' && inputParam !== null && typeof inputParam.value === 'string') {
    return inputParam.value.trim();
  } else if (typeof inputParam === 'string') {
    return inputParam.trim();
  }
  return inputParam;
}

function isEmptyValue(value) {
  return value == null || value === '';
}

function isVecType(type) {
  return type.indexOf('Vec<') >= 0;
}

function convertVector(value, type) {
  const converted = value.split(',').map((e) => e.trim());
  return converted.map((single) =>
    isNumType(type) ? (single.indexOf('.') >= 0 ? parseFloat(single) : parseInt(single)) : single
  );
}

function shouldConvertToBalance(chainInfo, converted) {
  if (!chainInfo?.tokenDecimals) return false;
  if (!converted.match(/^[0-9]+([.][0-9]+)?$/)) {
    onSetErrorMessage('Only numbers and "." are accepted.');
    return false;
  }
  const bnResult = inputToBn(chainInfo.tokenDecimals, converted);
  if (bnResult.gte(getGlobalMaxValue(128))) {
    onSetErrorMessage('Amount too large');
    return false;
  }
  return true;
}

function handleBalanceConversion(previousValue, converted) {
  const bnResult = inputToBn(chainInfo.tokenDecimals, converted);
  return [...previousValue, bnResult.toString()];
}

function convertToNumber(converted) {
  return converted.indexOf('.') >= 0 ? parseFloat(converted) : parseInt(converted);
}

export default transformParams;
```

Lykhoyda (changes requested):
If it's possible please refactor the method transformParams. Both scenarios mentioned in the description work fine 👍

Tbaut:
The refactor is in, very inspired by your code, thank you very much. I chose to only extract to functions some logic when
- it was more than 1 line
- it was re-used

This means that things like for instance the Balance handling is still on several lines, because I didn't want to duplicate the `bnResult = inputToBn(chainInfo.tokenDecimals, converted);`, but it should be much easier to read now.

Lykhoyda (approved):

packages/ui/src/components/EasySetup/ManualExtrinsic.tsx
```diff
@@ -175,21 +224,19 @@ const ManualExtrinsic = ({
...
-        const raw = getTypeDef(instance.toRawType())
-        console.log('raw', raw)
-
-        arg.typeName.isSome &&
-          console.log('typeName.unwrap().toString()', arg.typeName.unwrap().toString())
+        // console.log('getTypeDef', getTypeDef(arg.type.toString()))
+        // const instance = api.registry.createType(arg.type as unknown as 'u32')
+        // console.log('instance', instance)
+        // const raw = getTypeDef(instance.toRawType())
+        // console.log('raw', raw)
```
Lykhoyda, line 234: can we maybe move it to the log method and comment it?

Tbaut: After offline discussion we're keeping this around for a bit.

packages/ui/src/components/EasySetup/ManualExtrinsic.tsx
```diff
@@ -175,21 +224,19 @@ const ManualExtrinsic = ({
     let paramFields: ParamField[] = []
     const metaArgs = api.tx[palletRpc][callable].meta.args
 
-    console.log('metaArgs', metaArgs)
+    // console.log('metaArgs', metaArgs)
```
Tbaut, line 227: removing an easy one
```suggestion
```

Issue #380, opened 2023-09-29
Show the value in token with the right decimals, rather than in plank


Tbaut:
Very similarly to https://github.com/ChainSafe/Multix/issues/379
We should help our users review the Txs with the amounts converted in the right token, with the right decimals rather than showing the amount in plank. This is currently done in a somewhat hacky way only for `balances.transfer` alike here https://github.com/ChainSafe/Multix/blob/8711fb0ccbd67bdd1451b8024d64092de1a0db50/packages/ui/src/components/CallInfo.tsx#L35-L36

e.g this should display 123 ROC rather than 123,000,000,000,000
![image](https://github.com/ChainSafe/Multix/assets/33178835/5ea36ea1-7089-4c67-bfd2-62f7b1f74d58)

This means programmatically identifying when the amount is a balance (I belive the typeName is `BalanceOf`) and then convert to the amount of token, and add the token symbol.

Although this should be checked on Polkadot/Kusama/Rococo first, please check on other networks such as Hydra or Phala. The type names may be different because these things evolve.

Issue #134, opened 2023-05-10
Improve the readability of batch proposal review


asnaith:
With the addition of batch transactions, we now have more data to contend with and display on the proposal. 

It would be good to improve the readability. For example, this is a proposal in the transaction queue for a batch transaction of two balance transfers: 

<img width="568" alt="Screenshot 2023-05-10 at 4 32 40 PM" src="https://github.com/ChainSafe/Multix/assets/3982665/32abed94-3d8c-40ab-ad13-a2926a27a375">

Tbaut:
Any suggestion on what would make it more readable? like the address on one line or with the usual `AccountDisplay` anything else?

The hard part is that this is generic. so unless there's a specific rule (that's the case for transfers outside of batches) the formatting is generic as well and it may be tricky to handle well all cases. This is yet again an area where I'd like to have real users feedback. If someone is doing batches of transfer regularly, they'd benefit from it. I know the ChaosDao is doing batch referenda voting for instance, but not sure about transfers.

asnaith:
@Tbaut Yeah with it being generic and covering so many different parameters I can see the difficulty. 

That said, without the indentation, it would probably fit nicer in the space that's there, and if we styled each extrinsic with an outline or something it would be easy to see the separation of different extrinsic in the batch. 

The call, method, section and args will probably be present on all batch extrinsics. Perhaps we could display call, method and section data separately and put the "args" content in the main body? 

I created an example of what I mean (I'm no designer though 😄). 

<img width="568" alt="tx_queue" src="https://github.com/ChainSafe/Multix/assets/3982665/747714b9-915d-4f53-b8cb-b33148227d1b">

Tbaut:
I edited the name of this issue to reflect what can actually be done.
Your design adds 2 things
- style better the batch extrinsic (by separating the extrinsics, removing the 0, 1.
- style better the transfer if it's part of an extrinsic (right now, only stand alone transfers are formated nicely. This is a little less prio because very specific.

It's certainly doable though, the first one should definitely be the priority.

Tbaut:
I will close this for now, as I think things have improved a lot already. the overall styling could be improved, but that's not only for batch.

PR #384, opened 2023-10-04
Generically show token value without decimals

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/380
addresses a bit https://github.com/ChainSafe/Multix/issues/134

You can test with either a done transaction or by creating a new TX from callData, and looking at the decoded call, since the component is the same.

e.g with callData: `0x23000b00b04e2bde6f24736f6d652064657363`

image

There's also a huge improvement for batch, e.g callData `0x18000800002c736f6d652072656d61726b12000b00b04e2bde6f00f2c116d8c1f412f2dfc91702660acba3c4e9d487de4f7ac7577cb117264dd9aa`

image

it used to be displayed like this
image

### DIFF
--- a/packages/ui/src/components/CallInfo.tsx
+++ b/packages/ui/src/components/CallInfo.tsx
@@ -187,6 +187,7 @@ const CallInfo = ({
     () => aggregatedData.callData && getDecodeUrl(aggregatedData.callData),
     [aggregatedData, getDecodeUrl]
   )
+  const hasArgs = useMemo(() => args && Object.keys(args).length > 0, [args])
 
   return (
     <div className={className}>
@@ -211,7 +212,7 @@ const CallInfo = ({
           annoyance.
         </AlertStyled>
       )}
-      {args && !!api && Object.keys(args).length > 0 && (
+      {!!api && hasArgs && (
         <Expander
           expanded={expanded}
           title="Params"
--- a/packages/ui/src/components/CallInfo.tsx
+++ b/packages/ui/src/components/CallInfo.tsx
@@ -11,6 +11,8 @@ import { HiOutlineArrowTopRightOnSquare as LaunchIcon } from 'react-icons/hi2'
 import { Link } from './library'
 import { usePjsLinks } from '../hooks/usePjsLinks'
 import { Alert } from '@mui/material'
+import { ApiPromise } from '@polkadot/api'
+import { isTypeBalance } from '../utils/isTypeBalance'
 
 interface Props {
   aggregatedData: Omit<AggregatedData, 'from' | 'timestamp'>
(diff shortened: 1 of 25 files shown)


asnaith:
@Tbaut This is a really nice improvement, makes a lot more sense with the token value, and the display is much improved with the inclusion of the identicon of the receiver too. 

However, one issue I noticed is that "ROC" is being appended in a place it shouldn't be:

CleanShot 2023-10-06 at 10 06 47@2x

Tbaut:
You must have been testing an older commit could it be? The latest refractor fixes this bug that I found earlier.

asnaith (approved):
> You must have been testing an older commit could it be? The latest refractor fixes this bug that I found earlier.

I double-checked to make sure and then see what the confusion was. 

I had clicked the most recent "View Deployment" button and I was testing on there and still seeing the issue (on https://6ddde8e1.multix.pages.dev/?network=rococo). 

However, when I pulled the latest changes and tested on localhost I saw that it was fixed. Seems like there wasn't a new staging deployment for the very latest commits? Not sure why. 

This PR looks good to go though, no further concerns :)

packages/ui/src/components/CallInfo.tsx
```diff
@@ -140,11 +211,11 @@ const CallInfo = ({
           annoyance.
         </AlertStyled>
       )}
-      {args && Object.keys(args).length > 0 && (
+      {args && !!api && Object.keys(args).length > 0 && (
```
Lykhoyda, line 214: I really like it when we encapsulate such logic in the method. There is no super complex logic, but it's still hard to read. What about making the method:
```js
const containsArgs = () => args && Object.keys(args).length > 0

{containsArgs() && !!api (<div></div>)}
```

Also the variable casting for "api" is not 100 percent right as it could be an empty object which is returns true, it doesn't give us confidence that API is initialized. 

I personally use lodash for such scenarios to not create new methods or check for the emptiness of the objects (https://lodash.com/docs/4.17.15#isEmpty)

Lykhoyda (changes requested):
Please create a method for verifying the args data

packages/ui/src/components/CallInfo.tsx
Tbaut: Lodash is hiding things behind a giant lib that has had many vulns in the past, it's a strong no for me. These are tastes after all. I don't like to hide things too much when they're a 1 line, and if they're not reused.. I've been involved in such projects and it's easy to read, but hard to maintain. I'll implement this suggestion, but I don't want this kind of comment to be a requirement for this repo, or to prevent merging after being up for 6 days.

Lykhoyda (approved):

PR #385, opened 2023-10-05
feat: Update network select

### LINKED ISSUE #264: Rework the network dropdown

Lykhoyda:
Closes #264 
- Updated the network selection for the desktop
- Updated the network selection for mobile
- Mobile menu style updates
- fix "watch account" or "connect" for mobile devices
- fix loader appearance for MultiProxy selection when there is no proxy

### DIFF
--- a/packages/ui/src/components/select/MultiProxySelection.tsx
+++ b/packages/ui/src/components/select/MultiProxySelection.tsx
@@ -117,16 +117,16 @@ const MultiProxySelection = ({ className }: Props) => {
     />
   )
 
-  if (multiProxyList.length === 0) {
-    return null
-  }
-
   if (isLoading) {
     return (
       <BoxStyled>
         <CircularProgress size={24} />
       </BoxStyled>
     )
+  }
+
+  if (multiProxyList.length === 0) {
+    return null
   }
 
   return (
--- a/packages/ui/src/components/select/MultiProxySelection.tsx
+++ b/packages/ui/src/components/select/MultiProxySelection.tsx
@@ -117,16 +117,16 @@ const MultiProxySelection = ({ className }: Props) => {
     />
   )
 
+  if (multiProxyList.length === 0) {
+    return null
+  }
+
   if (isLoading) {
     return (
       <BoxStyled>
         <CircularProgress size={24} />
       </BoxStyled>
     )
-  }
-
-  if (multiProxyList.length === 0) {
-    return null
   }
 
   return (
@@ -155,16 +155,18 @@ const BoxStyled = styled(Box)`
 `
 
 export default styled(MultiProxySelection)`
-  min-width: 200px;
-
-  @media (min-width: ${(props) => props.theme.breakpoints.values.md}px) {
-    min-width: 260px;
-  }
-
+  min-width: 12.5rem;
   text-align: right;
+  width: 100%;
 
   .MuiTextField-root {
-    max-width: 18.875rem;
+    @media (min-width: ${(props) => props.theme.breakpoints.values.md}px) {
+      min-width: 13.6875rem;
+    }
+
+    @media (min-width: ${(props) => props.theme.breakpoints.values.lg}px) {
+      min-width: 18.6875rem;
+    }
   }
 
   .MuiAutocomplete-endAdornment {
--- a/packages/ui/src/components/select/NetworkSelection.tsx
+++ b/packages/ui/src/components/select/NetworkSelection.tsx
@@ -60,7 +60,7 @@ const NetworkSelection = () => {
     >
       <ListSubheader>Polkadot & Parachains</ListSubheader>
       {renderNetworks(polkadotNetworksAndParachains)}
-      <ListSubheader>Kusama & Parachainss</ListSubheader>
+      <ListSubheader>Kusama & Parachains</ListSubheader>
       {renderNetworks(kusamaNetworksAndParachains)}
       <ListSubheader>Testnets</ListSubheader>
       {renderNetworks(testChains)}
@@ -71,8 +71,16 @@ const NetworkSelection = () => {
 const MenuPropsStyles = {
   sx: {
     marginTop: '.75rem',
+
     '.MuiPaper-root': {
-      boxShadow: 'none'
+      marginTop: '2px',
+      boxShadow: 'none',
+      border: 'none',
+      outline: `1px solid ${theme.custom.text.borderColor}`,
+
+      ul: {
+        border: 'none'
+      }
     },
 
     '.MuiMenuItem-root': {
@@ -81,7 +89,7 @@ const MenuPropsStyles = {
     },
 
     '.MuiList-root': {
-      columns: '150px 2',
+      columnFill: 'balance',
       padding: 0,
       border: `1px solid ${theme.custom.text.borderColor}`,
       borderRadius: '0.5rem'
@@ -89,6 +97,24 @@ const MenuPropsStyles = {
 
     '.MuiListSubheader-root': {
       columnSpan: 'all'
+    },
+
+    [theme.breakpoints.up('xs')]: {
+      '.MuiList-root': {
+        columns: '160px 2'
+      }
+    },
+
+    [theme.breakpoints.up('sm')]: {
+      '.MuiList-root': {
+        columns: '180px 2'
+      }
+    },
+
+    [theme.breakpoints.up('md')]: {
+      '.MuiList-root': {
+        columns: '200px 2'
+      }
     }
   }
 }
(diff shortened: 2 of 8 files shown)


asnaith (changes requested):
Hey Anton, I noticed a couple of things here and also need a bit of clarification on a few things.

---
1. The network selector is less round and has a bit more space than the version on the issue / in figma. For example: 

CleanShot 2023-10-05 at 15 55 07@2x

---

2. On the mobile menu we are still including the network name and the connect button and menu items are centered and not right aligned. 

CleanShot 2023-10-05 at 15 43 14@2x

---
3. I noticed there's some gaps in the corners of the menu border:

CleanShot 2023-10-05 at 15 23 26@2x

---

4. It was not obvious to me what was fixed with regard to these two points below, could you please advise me on the nature of the change so that I can validate? 🙏
> fix "watch account" or "connect" for mobile devices
> fix loader appearance for MultiProxy selection when there is no proxy

---
 5. Is the network selection supposed to look like this on mobile? 

CleanShot 2023-10-05 at 16 02 22@2x

---

6. Lastly (and this could be out of the scope of this ticket) I feel like we should probably use a more distinct colour for the connect button now that we no longer have a blue background, what do you think? cc @Tbaut 

CleanShot 2023-10-05 at 16 08 58@2x

Tbaut:
For 6. I'd say I've no strong opinion. We could have it primary. But this doesn't have to be tackled in this PR since it's not par of it.

Lykhoyda:
> Hey Anton, I noticed a couple of things here and also need a bit of clarification on a few things.
> 
> 1. The network selector is less round and has a bit more space than the version on the issue / in figma. For example:
> 
> CleanShot 2023-10-05 at 15 55 07@2x
> 2. On the mobile menu we are still including the network name and the connect button and menu items are centered and not right aligned.
> 
> CleanShot 2023-10-05 at 15 43 14@2x
> 3. I noticed there's some gaps in the corners of the menu border:
> 
> CleanShot 2023-10-05 at 15 23 26@2x
> 4. It was not obvious to me what was fixed with regard to these two points below, could you please advise me on the nature of the change so that I can validate? 🙏
> 
> > fix "watch account" or "connect" for mobile devices
> > fix loader appearance for MultiProxy selection when there is no proxy
> 
> 5. Is the network selection supposed to look like this on mobile?
> 
> CleanShot 2023-10-05 at 16 02 22@2x
> 6. Lastly (and this could be out of the scope of this ticket) I feel like we should probably use a more distinct colour for the connect button now that we no longer have a blue background, what do you think? cc @Tbaut
> 
> CleanShot 2023-10-05 at 16 08 58@2x

@asnaith Thank you for the review!
1) The reason why the selector is not rounded is that the changes should be reflected in the other components as well. Because if we round only the network selector it would look strange. I think it's a recent change in the design to have more rounded components. So it can be changed in a different story as it will touch other components. For consistency, I would keep a similar design for the current PR.

Screenshot 2023-10-06 at 11 02 34

2) For the mobile menu I keep the name as we don't have a lack of space compared to the desktop version. It's also helpful to distinguish when the icon is the same but has different networks (Assethub-Dot and Assethub-Ksm). IMO it looks better when the items are centered, as your eyes are focused more on the middle of the screen and the text looks more readable. In the newer version of the mobile design the network selector and proxy are not part of the menu, so it will be changed anyway :) 

3) Good catch!

4) 
fix "watch account" or "connect" for mobile devices
Screenshot 2023-10-06 at 11 17 54

---- 
"fix loader appearance for MultiProxy selection when there is no proxy"

When you switch the network and have no multisig we show the loader and it disappears after checking that there is no multisig. It causes the UI to unnecessarily "jump".

Screenshot 2023-10-06 at 11 19 28

5) So far we don't have a mobile design. I will make it a little bit bigger, but yeah it's how it looks now.

packages/ui/src/constants.ts
```diff
@@ -189,6 +189,19 @@ export const networkList = {
...
+  'hydradx',
+  'astar',
+  'moonbeam',
+  'assethub-dot',
+  'bifrost-dot',
+  'interlay',
+  'phala'
+]
+export const kusamaNetworksAndParachains = ['kusama', 'khala', 'assethub-ksm', 'moonriver']
+export const testChains = ['rococo', 'acala', 'rhala testnet', 'local']
```
Tbaut, line 203: acala is a Polkadot chain :)
For the order, I'd suggest to have the relay-chain and asset-hub first (system parachain) the rest in alphabetical order.

Tbaut (commented):
Some comments on one decision regarding the loader. I don't have strong opinions regarding what was discussed above with Andrew.

packages/ui/src/constants.ts
Lykhoyda: Yeah, I need help here to validate if everything is correct

Lykhoyda: @Tbaut is it a correct order?
```js
export const polkadotNetworksAndParachains = [
  'polkadot',
  'acala',
  'moonbeam',
  'hydradx',
  'phala',
  'bifrost-dot',
  'assethub-dot',
  'astar',
  'interlay'
]
export const kusamaNetworksAndParachains = ['kusama', 'khala', 'moonriver', 'assethub-ksm']
export const testChains = ['rococo', 'rhala testnet', 'local']
```

Lykhoyda: BTW why we don't have a westend? I thought it was a popular test net.  😃

packages/ui/src/components/select/MultiProxySelection.tsx
```diff
@@ -117,6 +117,10 @@ const MultiProxySelection = ({ className }: Props) => {
     />
   )
 
+  if (multiProxyList.length === 0) {
```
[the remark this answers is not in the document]
Lykhoyda: okay, let's solve it separately

packages/ui/src/constants.ts
Tbaut: There was no subsquid archive available for Westend (those are the archives supplied by subsquid) and although they said it's planned, it didn't happen yet. So it means that we don't have any indexer ready for it.

Alphabetical order starting after the relay + system chains should make it so
```ts
export const polkadotNetworksAndParachains = [
  'polkadot',
  'assethub-dot',
  'acala',
  'astar',
  'bifrost-dot',
  'hydradx',
  'interlay'
  'moonbeam',
  'phala',
]
export const kusamaNetworksAndParachains = ['kusama', 'assethub-ksm', 'khala', 'moonriver']
export const testChains = ['rococo', 'rhala testnet', 'local']

File: packages/ui/src/components/select/NetworkSelection.tsx [hunk 3]
```diff
@@ -14,31 +21,179 @@ const NetworkSelection = () => {
...
+  max-width: 9.1875rem;
+  box-sizing: content-box;
+`
+
+const ImgStyled = styled('img')`
+  width: 1.25rem;
+  height: 1.25rem;
+  margin-right: 0.5rem;
+  border-radius: 50%;
+  max-height: 2rem;
```
Tbaut, line 189: Should we have a much bigger logo then? I've had nice results from a `2rem` and some padding changes on the `.MuiSelect-select` see next comment.
```suggestion
  width: 2rem;
  height: 2rem;
  margin-right: 0.5rem;
  border-radius: 50%;
```

packages/ui/src/components/select/NetworkSelection.tsx
```diff
@@ -14,31 +21,179 @@ const NetworkSelection = () => {
...
+
+  @media (min-width: ${({ theme }) => theme.breakpoints.values.md}px) {
+    margin-left: 1rem;
+    max-width: fit-content;
+  }
+
+  .MuiSelect-select {
+    display: flex;
+    padding-right: 2.5rem !important;
+
```
Tbaut, line 139: reducing the padding here to fit for a bigger logo
```suggestion
padding-top: 0.75rem;
padding-bottom: 0.75rem;

```

Tbaut (approved):
Looking good from my side. I think the logo could be bigger in the network menu see my comments below. I opened an issue for the jumping annoyance https://github.com/ChainSafe/Multix/issues/387.

packages/ui/src/components/select/NetworkSelection.tsx
Lykhoyda: thank you for the suggestion! 2 rem is probably too big, 1.75 looks better for me

PR #386, opened 2023-10-05
Add pr submission template

asnaith:
Hey @Tbaut, @Lykhoyda 

I put together this PR template. As mentioned in the retro, the idea is to give us a reminder at the submission stage of things we may otherwise overlook. 

Please feel free to suggest changes and add / remove any items you think are needed or unnecessary 🙏

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,7 +11,7 @@ Submission checklist:
 
 #### Compatibility
 - [ ] Functionality of change validated with a connected account with multisig
-- [ ] Applicable elements hidden / disabled for watched accounts
+- [ ] Applicable elements hidden / disabled for watched multisigs / pure
 
 #### Maintenance
 - [ ] Tests added 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,23 +1,24 @@
 closes #
 
 ---
-Submission checklist: 
 
-<!-- Remove anything below that is not applicable -->   
+Submission checklist:
+
+<!-- Remove anything below that is not applicable -->
 
 #### Layout
+
 - [ ] Change inspected in the desktop web ui
 - [ ] Change inspected in the mobile web ui
 
 #### Compatibility
+
 - [ ] Functionality of change validated with a connected account with multisig
 - [ ] Applicable elements hidden / disabled for watched multisigs / pure
 - [ ] Looks good for solo multisig
 - [ ] Looks good for multisig with proxy
 
 #### Maintenance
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+Submission checklist: 
+
+<!-- Remove anything below that is not applicable -->   
+
+#### Layout
+- [ ] Change inspected in the desktop web ui
+- [ ] Change inspected in the mobile web ui
+
+#### Compatibility
+- [ ] Functionality of change validated with a connected multisig
```
Tbaut, line 13:
```suggestion
- [ ] Functionality of change validated with a connected account with multisig
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+- [ ] Applicable elements hidden / disabled for watched accounts
```
Tbaut, line 14:
```suggestion
- [ ] Applicable elements hidden / disabled for watched multisigs / pure
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+
```
Tbaut, line 15: I'd also add the solo multisig and multisig with pure.
```suggestion
- [ ] Looks good for solo multisig
- [ ] Looks good for multisig with proxy
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+#### Maintenance
+- [ ] Tests added 
+- [ ] Tests updated
+- [ ] Issue for test coverage created
+- [ ] Documentation added
+- [ ] Documentation updated
```
Tbaut, line 21: While it's probably good to have tests etc at the same time, I would say it's too early for addind test at the same time.
with added/updated those are either or afaiu, where I think we should go for all checkboxes checked as a visual sign to be able to merge.

.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
+closes #
+
+---
+
+Submission checklist:
+
+<!-- Remove anything below that is not applicable -->
+
```
Tbaut, line 8: Either uncomment this line or remove, because it's not visible. LGTM otherwise.
I think the documentation is also a bit overkill. Like we never have doc :dancers:

Tbaut (approved):

.github/pull_request_template.md
asnaith: @Tbaut ok, removed :)

PR #388, opened 2023-10-06
Small refactor to prevent casting selected value.

Tbaut:
Came across the `as SelectChangeEvent<string>` in the code and thought we could do better. It ended up still adding more code than what we had, but it's a little nicer and more generic I think. wdyt @Lykhoyda?


Lykhoyda (approved):
Looks good! thank you!

PR #390, opened 2023-10-06
Prevent "no multisig" from being displayed before we checked

### LINKED ISSUE #381: Message "no multisig found for your accounts" showed briefly on refresh although there are multisigs
**Description**
When hitting refresh, a loaders appears to connect to the rpc node, then there's a message displayed briefly saying that I don't have any multisig for my account, although it's not true. Then a loader saying it's fetching my multisigs, then the expected multisigs then do appear.

Ideally, the message saying I've no multisig for my accounts should not be displayed.

**Steps To Reproduce**

Using Chrome as my browser
Just hit refresh

https://github.com/ChainSafe/Multix/assets/75752467/03d4af2a-7166-42d0-9a09-5d87d6923a91

Tbaut:
closes #381 

The flow of things was as follow
- connect to a node
- check whether this is an ethereum network we're connecting to
- --> here we showed the "no multisig" before even checking
- depending on the type of network, we request specific accounts (of type substrate or ethereum)
- query the indexer for known multisigs

There are other flashing things like the "loading multisig" but it's a little more tricky to fix.
- fixed some nits like stray console.log
- prevent walletconnect from spamming the console

Submission checklist:

#### Layout


Tbaut:
Deployment is failing because of Cloudflare issues. The reviewing of this PR shouldn't be impacted though.

Lykhoyda (approved):

Tbaut:
Found a bug, let's not merge.

Tbaut:
The bug was visible when you had a watched account with no multisig, but no extension connection. The spinner would show endlessly.

I changed a little my strategy. Making sure the first "Mutlxi is an interface.." would only show when we're sure we've init all the localstorage related elements (connection to the extension and watched accounts). 
The "no multisig" is still displayed, due to the way our state is handled and loosely coupled. But we're now talking about a couple ms, which is less than the blink of an eye, as I think a big refactor to prevent this isn't worth it rn.

packages/ui/cypress/support/commands.ts
```diff
@@ -55,7 +55,7 @@ Cypress.Commands.add('initExtension', (accounts: InjectedAccountWitMnemonic[]) =
 })
 
 Cypress.Commands.add('getAuthRequests', () => {
-  return cy.wrap(extension.getAuthRequests())
+  return cy.wait(500).then(() => cy.wrap(extension.getAuthRequests()))
```
Lykhoyda, line 58: We can use waitUntil especially when we show the loader. It's hard to guess how much time to really need, so the test might be unstable.
The example of usage
```
 cy.waitUntil(function() {
  return cy.get('loader').should('not.exist');
 })

Comment (Lykhoyda) [reply packages/ui/cypress/support/commands.ts:58]: There is a library for that method https://www.npmjs.com/package/cypress-wait-until

It's a missing method in the cypress, but it's inspired by "Testing library" (https://testing-library.com/docs/guide-disappearance/#waiting-for-disappearance).

File: packages/ui/src/pages/Home/Home.tsx [hunk 2]
```diff
@@ -93,30 +107,27 @@ const Home = ({ className }: HomeProps) => {
...
-          ? 'Loading accounts...'
-          : `Connecting to the node at ${selectedNetworkInfo?.rpcUrl}`}
-      </Box>
+        Loading accounts...
+      </LoaderBoxStyled>
     )
   }
 
-  if (isLoading) {
+  if (isLoadingMultisigs) {
```
Lykhoyda, line 130: What about to move the loading and error handling in the separate file:
```jsx
import React from 'react';
import { CircularProgress } from '@mui/material';
import { LoaderBoxStyled, MessageWrapper, ErrorMessageStyled } from './YourStyledComponents'; // Update with your actual imports

const useLoadingAndError = ({ api, isAccountLoading, isLoadingMultisigs, multisigQueryError, multiProxyList, showNewMultisigAlert, selectedNetworkInfo }) => {
  if (!api) {
    return (
      <LoaderBoxStyled data-cy="loader-rpc-connection">
        <CircularProgress />
        {`Connecting to the node at ${selectedNetworkInfo?.rpcUrl}`}
      </LoaderBoxStyled>
    );
  }

  if (isAccountLoading) {
    return (
      <LoaderBoxStyled data-cy="loader-accounts-connection">
        <CircularProgress />
        Loading accounts...
      </LoaderBoxStyled>
    );
  }

  if (isLoadingMultisigs) {
    return (
      <MessageWrapper>
        <CircularProgress />
        <div>Loading your multisigs...</div>
      </MessageWrapper>
    );
  }

  if (multisigQueryError) {
    return (
      <MessageWrapper>
        <ErrorMessageStyled>
          <ErrorOutlineIcon size={64} />
          <div>An error occurred.</div>
        </ErrorMessageStyled>
      </MessageWrapper>
    );
  }

  if (multiProxyList.length === 0) {
    return (
      <MessageWrapper>
        {showNewMultisigAlert ? <SuccessCreation /> : <ConnectOrWatch />}
      </MessageWrapper>
    );
  }

  return null; 
};

export default useLoadingAndError;
```

Home page
```jsx
  const loadingAndErrorComponent = useLoadingAndError({
    api,
    isAccountLoading,
    isLoadingMultisigs,
    multisigQueryError,
    multiProxyList,
    showNewMultisigAlert,
    selectedNetworkInfo,
  });

  if (loadingAndErrorComponent) {
    return loadingAndErrorComponent;
  }
  ```

Lykhoyda (changes requested):
Please check the refactoring suggestion. Let me know if you want to work on it in the separate issue

packages/ui/src/pages/Home/Home.tsx
Tbaut: Sure, that'd help make the Home easier to read, I'll do it in a normal component though, rather than a hook.

packages/ui/cypress/support/commands.ts
Tbaut: I really dislike `waits`, and I'll fight them all as much as I can. But in this case, I don't think we have any other solution, because the actual firing of the request is itself delayed. the loader appears right away (after this PR) but the firing of the request for accounts is done 500ms after:
https://github.com/ChainSafe/Multix/blob/eac29c0b55c704a88ace236e5ad98af02f1cea34/packages/ui/src/contexts/AccountsContext.tsx#L136-L138

Tbaut: I'll check if waitUntil can help here

packages/ui/src/pages/Home/Home.tsx
Tbaut: I pushed an update separating `DisplayError` and `DisplayLoader`, and keeping the rest as is for now.

Tbaut:
All right, `waitUntil` is much nicer. I want to keep it out of the cypress commands though so as not to request more deps from our users when the plugin is deployed out of Multix

Lykhoyda (approved):
Looks good! Thank you for the updates

asnaith (approved):

PR #392, opened 2023-10-09
Generic accounts displayed in call info

### LINKED ISSUE #379: Show address with identicon in Tx review when the param type is an address
For the review of manual extrinsics, we show the address plainly, without the identicon or a name, e.g
image

There's currently a hack to show it nicely only for `balances.transfer` and `proxy.addProxy` I believe.

We should enhance this and check the type of params generically, and show a `<AccountDisplay/>` which will show the name if any/identity and the identicon.

I didn't do the research to know exactly what types are linked to such fields `AccountId`, or maybe it `AccountIdLookupOf` or `MultiAddress`?

Tbaut:
closes #379 

Similar to https://github.com/ChainSafe/Multix/pull/384 but for addresses and multisigs.
We now will show the identicon, and eventually the name of the address if stored either locally, or in the extension.

The testing is similar as well, you can create some extrinsics on pjs/apps and see the result of the decoding in the new tx > from call data. E.g with batch

image

---

Submission checklist: not applicable

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -31,7 +31,7 @@ If possible, please copy/paste or provide a screenshot of the browser console. T
 **Screenshots / Video**
 If applicable, please add screenshots or video to help explain or demonstrate the problem you are experiencing.
 
-**Test Host Specifics: **
+**Test Host Specifics**
 
 - OS: [e.g. Linux Mint 21.1, macOS 13, Windows 11]
 - Browser [e.g. Chrome, Safari]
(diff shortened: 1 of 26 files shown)


Lykhoyda:
@Tbaut It works well! I have only concerns about how it looks, the data appearance is pretty dense. We need to consider the design for the transaction list improvement story

Screenshot 2023-10-09 at 16 48 12

Tbaut:
Yes, this issue https://github.com/ChainSafe/Multix/issues/134 has evolved a bit over time, and we're already in a better state than where we were. The main issue is to show things nicely, but in a maintainable and generic way. Unordered list were acceptable so far, but we need to do better eventually

Lykhoyda (approved):

Tbaut:
I'll wait for https://github.com/ChainSafe/Multix/pull/384 to be reviewed before merging, since it's be harder to review otherwise.

PR #393, opened 2023-10-10
Move to Arrow-squid

### LINKED ISSUE #372: Bump indexer deps and migrate to Arrow-squid
The current version of subsquid will be deprecated in December. We should move to arrow-squid asap. Migration steps https://docs.subsquid.io/migrate/migrate-to-arrowsquid-on-substrate/

although many networks aren't available yet, we should test and get ready.
Available networks: https://docs.subsquid.io/substrate-indexing/supported-networks/

Tbaut:
closes #372

- There should be no change in the interface, or actually anything. It's purely a backend upgrade that should be transparent.
- 90% of the diff is auto-generated here, such as everything in `squid/src/types`
- some RPCs from onFinality were unresponsive
- I merged the chopsticks PR into this one because Chopsticks needed Arrow-squid, and both PRs in parallel was more annoying than anything. Right now there's a config file for Chopstick that will be used when we close https://github.com/ChainSafe/Multix/issues/394


Tbaut:
I'm moving Joystream to RPC, takes a little while to sync, ~2h but there will be an archive in the next weeks.

squid/src/main.ts
```diff
@@ -70,19 +59,22 @@ const processor = new SubstrateBatchProcessor()
...
-  .addEvent('Proxy.PureCreated', dataEvent)
-  .addEvent('Proxy.AnonymousCreated', dataEvent)
-  .addEvent('Proxy.ProxyAdded', dataEvent)
-  .addEvent('Proxy.ProxyRemoved', dataEvent)
-
-export type Item = BatchProcessorItem<typeof processor>
-export type Ctx = BatchContext<Store, Item>
+  .setFields({ ...fields })
+  .addCall({
+    name: [
```
Lykhoyda, line 64: We can move it to the constant

Tbaut: totally. Done in [e8a2fde](https://github.com/ChainSafe/Multix/pull/393/commits/e8a2fdee272d99330a1ec517d0c904dca1c2092f)

squid/src/main.ts
```diff
@@ -29,39 +24,33 @@ import { Env } from './util/Env'
...
 
+export const fields = {
+  call: {
+    args: true,
+    origin: true,
+    success: true
+  },
+  event: {
+    args: true
+  },
```
Lykhoyda, line 60: Not every call will have an event data, For example for kill proxy event has no event

Tbaut: This is about the structure of events. if it's an event, I need the args. 
https://docs.subsquid.io/substrate-indexing/setup/field-selection/#set-fields

`proxy.killProxy` is a call, it'll have `args`, `origin`, `success`, which is what we use.

Lykhoyda (approved):
Looks good! 

Please consider the scenario when we have no event data if it works as expected

Tbaut:
I don't understand what you mean. see my answer to you comment.

Lykhoyda:
> I don't understand what you mean. see my answer to your comment.

It was before I saw your comment. All good 🙌

asnaith (approved):
The multisigs I checked were all viewable as expected 

Looks good to go :)

Issue #378, opened 2023-09-29
Write tests to validate watch accounts


Tbaut:
This suite of test should make sure that:

- [ ] we can add a watched account with an expected result (say it has x accounts/pure in the dropdown)
- [x] we can remove a watched account
- [ ] add either the proxy or a multisig with the same results
- [x] the name given while adding the account is reflected in the home and in the main dropdown
- [x] we can edit the name of a watched account (with 3 dots menu)
- [x] we can see the subscan link (in the 3 dots menu)
- [x] we can see the tx of a watched account
- [x] we can **not** sign it
- [x] we can **not** send a Tx with a watched account
- [x] we can **not** add a walletConnect connection

To make sure this remains consistent over time, I believe that we should create a throw away account, we should
- create a multisig without proxy
- create a multisig with proxy
- create a tx (to have a tx to watch) with either
- share the mnemonic in the PR in case we want to change something in the future, but ideally not touch this account any more.

PR #396, opened 2023-10-11
Additional watched account tests

asnaith:
related to #378

New tests for

- Adding custom names to watched multisigs / watched pures 
- Checking the correct badge is displayed for pures / multisigs
- Checking new transaction button does not appear for watched accounts
- Checking the subscan link is present on the options menu 

Also removed some unnecessary double quotes in the locators on the page objects


packages/ui/cypress/support/page-objects/components/accountDisplay.ts
```diff
@@ -0,0 +1,7 @@
+export const accountDisplay = {
```
asnaith, line 1: Decided to create this a test object itself as it appears all over the app and then when checking for it's presence we can get cypress to yield the parent object and then check 'within' it to make sure we are checking the right ones (as account display details could appear on a page in multiple places etc)

packages/ui/cypress/tests/watched-accounts.cy.ts
```diff
@@ -58,4 +63,154 @@ describe('Watched Accounts', () => {
...
+      accountDisplay.pureBadge().should('not.exist')
+    })
+    // navigate to the home page and ensure the multisig badge is included in the header
+    topMenuItems.homeButton().click()
+    multisigPage.accountHeader().within(() => {
+      accountDisplay.identicon().should('be.visible')
+      accountDisplay.multisigBadge().should('be.visible')
+      accountDisplay.pureBadge().should('not.exist')
+    })
+  })
```
asnaith, line 142: created the badge-checking tests separate from the name-checking tests, I thought it was best to keep them separate so that if they fail it's immediately clear. Agreed?

packages/ui/cypress/tests/watched-accounts.cy.ts
```diff
@@ -58,4 +63,154 @@ describe('Watched Accounts', () => {
...
+    addWatchAccount(
+      watchMultisigs['multisig-with-pure'].pureAddress,
+      watchMultisigs['multisig-with-pure'].name)
+    // navigate to the home page and edit the name
+    topMenuItems.homeButton().click()
+    multisigPage.optionsMenuButton().click()
+    multisigPage.editNamesMenuOption().click()
+    editNamesModal.body().should('be.visible')
+    editNamesModal.inputEditPureName().type(`{selectall}{del}${`Edited Name Test`}`)
+    cy.wait(500)
```
asnaith, line 178: This is another annoying situation where this 500ms wait is needed before clicking the save button, without it the test will fail because the name in the header never updates. With this slight wait, it works every time. 

I understand we never want to never use waits but this is another instance where I couldn't go further without one. Open to suggestions.

packages/ui/cypress/tests/watched-accounts.cy.ts
```diff
@@ -58,4 +63,154 @@ describe('Watched Accounts', () => {
...
+  it('can see a subscan link for a watched pure', () => {
+    cy.visit(settingsPageWatchAccountUrl)
+    addWatchAccount(
+      watchMultisigs['multisig-with-pure'].pureAddress,
+      watchMultisigs['multisig-with-pure'].name)
+    // navigate to the home page and edit the name
+    topMenuItems.homeButton().click()
+    multisigPage.optionsMenuButton().click()
+    multisigPage.subscanMenuOption().should('be.visible')
+  })
```
asnaith, line 204: This is a very basic test atm, do we want to try and remove the target on the link so that the tab loads in the same window and then check that the subscan URL contains the correct address....or is that too much?

packages/ui/src/components/AccountEditName.tsx
```diff
@@ -66,6 +67,7 @@ const AccountEditName = ({ address, onNameChange, className }: Props) => {
           onChange={onChange}
           disabled={isExtensionAccount}
           value={name}
+          data-cy={`input-edit-${testId}-name`}
```
asnaith, line 70: even though passing the testId as a prop I'm still using data-cy as it's an internal component, I believe that's what we said to do in the past (Only use data-testid when it's an external component).

packages/ui/src/components/Header/Header.tsx
```diff
@@ -42,6 +42,7 @@ const Header = ({ handleDrawerOpen }: Props) => {
                 <NavLink
                   key={name}
                   to={path}
+                  data-cy={`button-navigate-${name.toLowerCase().replace(/ /g, '-')}`}
```
asnaith, line 45: Ok with this? We're making the links programmatically so think I have to do the same for the test tags and doing this means we can have consistent looking locators eg

`editNamesMenuOption: () => cy.get('[data-cy=menu-option-edit-names]')`

instead of 

  `editNamesMenuOption: () => cy.get('[data-cy=menu-option-Edit names]')`

asnaith:
@Tbaut @Lykhoyda I left some comments on specific things here, wanted to seek opinions before going further with more tests.

packages/ui/cypress/tests/watched-accounts.cy.ts
```diff
@@ -58,4 +63,146 @@ describe('Watched Accounts', () => {
...
+      accountDisplay.nameLabel().should('have.text', watchMultisigs['multisig-with-pure'].name)
+    })
+    // navigate to the multisig page and ensure the name is included in the home page header
+    topMenuItems.homeButton().click()
+    multisigPage.accountHeader().within(() => {
+      accountDisplay.nameLabel().should('have.text', watchMultisigs['multisig-with-pure'].name)
+    })
+  })
+
+  it('can see the identicon and badge displayed for a watched multisig', () => {
```
Tbaut, line 117: I feel like this test could be put together with the one before. Because 90% is the same.

packages/ui/cypress/tests/watched-accounts.cy.ts
Tbaut: Ha that was my comment above. To me, I think we could merge them,but I have no strong opinion. I feel like this is bloating minimally the tests, but is very much related to the previous checks.

packages/ui/cypress/tests/watched-accounts.cy.ts
Tbaut: Let's see what we can do here. It's weird that the input are so sensitive. Maybe we can wait until the text is visible in the input or something. I'll check.

packages/ui/cypress/tests/watched-accounts.cy.ts
Tbaut: I imagine that you meant to write this for the subscan test. I agree that it'd be better to actually test that the link is for the right address at least. I didn't research it but I guess it's possible. Removing the target sounds like hack though :shrug:

packages/ui/src/components/Header/Header.tsx
Tbaut: Sure, sounds good.

packages/ui/src/components/IdenticonBadge.tsx
```diff
@@ -44,6 +44,7 @@ export const IdenticonBadge = ({
       color="primary"
       badgeContent={badge}
       anchorOrigin={{ horizontal: 'left', vertical: 'top' }}
+      data-cy={badge === AccountBadge.PURE ? 'badge-pure' : 'badge-multisig'}
```
Tbaut, line 47: I'd use the badgeContent variable here.
`badge-${badgeContent}`

packages/ui/cypress/tests/watched-accounts.cy.ts
asnaith: Merged :)

packages/ui/src/components/IdenticonBadge.tsx
asnaith: Updated to `data-cy={`badge-${badge}`}` 

(badgeContent property is using the badge prop)

packages/ui/cypress/tests/watched-accounts.cy.ts
Tbaut: Need to spend time on this, sorry that I didn't get to it today, will check asap and report back.

Tbaut: I learned something pretty :exploding_head: today with `cy.clock` and `cy.tick`. The reason you had to wait, is because of the way we handle the storage of names. We don't store names on press, we debounce it. Without de-bouncing this field, it appeared unresponsive (I remember adding this). There's a debounce of 300ms here https://github.com/ChainSafe/Multix/blob/2d4fa20d13af25b365fe327a480b8710ffcb4f94/packages/ui/src/components/AccountEditName.tsx#L30-L31

Now thanks to `cy.clock` and `tick` we can bend time..  and make things happen quicker :shrug: 
https://docs.cypress.io/api/commands/tick

Tbaut: @Lykhoyda I'll try to use this in the connect function instead of `waitUntil`. Because it is actually retrying stuff every 200ms, so it's not really elegant

asnaith: > I learned something pretty 🤯 today with `cy.clock` and `cy.tick`. The reason you had to wait, is because of the way we handle the storage of names. We don't store names on press, we debounce it. Without de-bouncing this field, it appeared unresponsive (I remember adding this). 

@Tbaut This is great! The explanation totally makes sense. Thank you for working this out, this is something I would have never considered but now we can look out for it in other scenarios! 🙏

(cc @juans-chainsafe - there's some good info in the comment above, we can look out for this in other frontend projects when encountering those scenarios where we think we need a `cy.wait`).

packages/ui/cypress/tests/watched-accounts.cy.ts
asnaith: Updated: @Tbaut Juan has a nice solution for checking external links on another FE project. The same technique works well for us

Tbaut: yup that's great, I figured there were some ways to stub this but didn't check :+1:

Tbaut (approved):
Wow that's amazing, and that's already a lot covered!

Tbaut:
@asnaith you didn't share with us the mnemonic of this accounts. Any way you can do it, or didn't you use a new account for this?

asnaith:
> @asnaith you didn't share with us the mnemonic of this accounts. Any way you can do it, or didn't you use a new account for this?

@Tbaut Yeah I had the signatories in this file
https://github.com/ChainSafe/Multix/blob/main/packages/ui/cypress/fixtures/watchAccounts/watchSignatories.ts

but your question made me check...and when I did I realized one of the mnemonics needed updating as it was incorrect, now all is correct in this pr: https://github.com/ChainSafe/Multix/pull/409

PR #398, opened 2023-10-12
Use clock & tick instead of waituntil

Tbaut:
Small refactor to prevent using `waitUntil` when we know there's a setTimeout.

### DIFF
--- a/packages/ui/cypress/utils/clickOnConnect.ts
+++ b/packages/ui/cypress/utils/clickOnConnect.ts
@@ -1,0 +1,11 @@
+import { landingPage } from '../support/page-objects/landingPage'
+import { topMenuItems } from '../support/page-objects/topMenuItems'
+
+export const clickOnConnect = () => {
+  cy.clock()
+  topMenuItems.connectButton().click()
+  landingPage.accountsLoader().should('contain', 'Loading accounts')
+  cy.tick(500)
+  // restore the clock to prevent WS errors
+  cy.clock().invoke('restore')
+}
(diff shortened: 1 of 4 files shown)


Tbaut:
Need to check again, somehow the clock/tick doesn't seem to work well here.

packages/ui/cypress/utils/clickOnConnect.ts
```diff
@@ -0,0 +1,11 @@
+import { landingPage } from '../support/page-objects/landingPage'
+import { topMenuItems } from '../support/page-objects/topMenuItems'
+
+export const clickOnConnect = () => {
+  cy.clock()
+  topMenuItems.connectButton().click()
+  landingPage.accountsLoader().should('contain', 'Loading accounts')
+  cy.tick(500)
```
Lykhoyda, line 8: As I understood from the documentation, it's a method to mock the time. I didn't find any setInterval in our implementation. Isn't it just pending Promise, or am I mistaken?

Tbaut: Nope there's a settimeout to go around a bug https://github.com/ChainSafe/Multix/blob/2a1f23bf2b9e5803a0fe7e3795e9c0f1c5579ed1/packages/ui/src/contexts/AccountsContext.tsx#L135-L138

Tbaut:
I'm giving up. The time i spent isn't worth it imho opened https://github.com/ChainSafe/Multix/pull/403 to do a small refactor and that's it

PR #399, opened 2023-10-12
Fix value is null

Tbaut:
Hotfix for a bug in prod... because of `any` :roll_eyes: 

With for instance `0x1a00181400c502008500042c0f5205000000000000000000001400c902000500042c0f5205000000000000000000001400cd02008500042c0f5205000000000000000000001400d102008500042c0f5205000000000000000000001400d502008500042c0f5205000000000000000000001400d902008500042c0f520500000000000000000000`

We had 
image


Tbaut:
merging as small and bug in prod.

PR #404, opened 2023-10-16
Easy flow to set an identity

### LINKED ISSUE #373: Add an easy tx flow to set an identity
This is a request by many new bounty curators. They want to setup a multisig with proxy.. and setting an identity is the next thing they are looking for.

we should:
- add a "set identity" in the dropdown for new tx
- allow setting it for a multisig or a pure
- show all the fields like "display name", "legal name"...
- make sure that those fields are already pre-filled in case there's already an identity set on-chain

Tbaut:
closes #373 

Adds an easy setup to set or modify and on-chain identity

image
image

---

Submission checklist:

#### Layout

#### Compatibility


asnaith:
This is a really nice improvement, working great for adding and updating, all working as expected. 

I found a crash in the case where an address is pasted into one of the inputs (see the video where I just copied from the identicon and then pasted). Pasting in normal strings works fine. 

https://github.com/ChainSafe/Multix/assets/3982665/4b5baef8-a2b9-4371-af16-a464c8dadf68

Tbaut:
Ah interresting, I didn't realize there was a limit of char. I'll implement this.

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
```diff
@@ -0,0 +1,127 @@
...
+    web: getRawOrNone(web),
+    riot: getRawOrNone(riot),
+    email: getRawOrNone(email),
+    pgpFingerprint: null,
+    image: getRawOrNone(image),
+    twitter: getRawOrNone(twitter)
+  }
+}
+
+const IdentitySetIdentity = ({ className, onSetExtrinsic, from }: Props) => {
```
Lykhoyda, line 49: we normally call Props by the name of the component, so it should be IdentitySetProps . Btw The name of the component is not looking right.

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const { api, chainInfo } = useApi()
+  const [identity, setIdentity] = useState<IdentityFields | undefined>()
```
Lykhoyda, line 51: Can we manage this state inside useIdentity?

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
```diff
@@ -0,0 +1,127 @@
...
+      setIdentity({
+        display: undefined,
+        legal: undefined,
+        web: undefined,
+        riot: undefined,
+        email: undefined,
+        image: undefined,
+        twitter: undefined
+      })
+    }
```
Lykhoyda, line 76: It could be also moved to useIdentity hook

packages/ui/src/components/modals/Send.tsx
```diff
@@ -20,19 +20,27 @@ import FromCallData from '../EasySetup/FromCallData'
...
 
-const SEND_TOKEN_MENU = 'Send tokens'
-const FROM_CALL_DATA_MENU = 'From call data'
-const MANUEL_EXTRINSIC_MENU = 'Manual extrinsic'
+export const easyTransferTitle = [
+  'Send tokens',
+  'From call data',
+  'Manual extrinsic',
+  'Set identity'
+] as const
```
Lykhoyda, line 30: Probably it's a case to enum:
enum EasyTransferTitle {
  SendTokens = 'Send tokens',
  FromCallData = 'From call data',
  ManualExtrinsic = 'Manual extrinsic',
  SetIdentity = 'Set identity',
}

packages/ui/src/components/modals/Send.tsx
```diff
@@ -119,24 +129,30 @@ const Send = ({ onClose, className, onSuccess, onFinalized }: Props) => {
     [getMultisigByAddress, selectedMultiProxy]
   )
 
-  const easySetupOptions: { [index: string]: ReactNode } = useMemo(() => {
+  const easySetupOptions: Record<EasyTransferTitle, ReactNode> = useMemo(() => {
     return {
-      [SEND_TOKEN_MENU]: (
+      'Send tokens': (
```
Lykhoyda, line 134: With enum, it will be easier to use the value without specifying a string

Lykhoyda (changes requested):
The UI works as expected, I was able to create a transaction and sign and my multisig name was changed. 

TBH for me as a quite new person to the Polkadot ecosystem the field data labels are confusing, it's hard to guess where all this data would be reflected. Can we at least add input placeholders to give more information about the data we ask?

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
Tbaut: This component is doing an `identity.setIdentity` call, hence the name of this component, so this isn't a mistake. But you're right that it doesn't look easy to read or doesn't follow any other component's naming. I'll change for `SetIdentity`.

> we normally call Props...

This is not true, a quick search proves that we have a lot of such `Props`. To me, this doesn't matter at all because it's not exported.

image

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
Tbaut: I don't understand what you mean. This is the state of the specific fields. This isn't directly related to the identity from the chain. I can change the name to `identityFields` if that's more clear.

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
Tbaut: Why? `useIdentity` is only there to fetch the identity from the chain. This is a hook that is re-used for other needs (the `AccountDisplay`)

packages/ui/src/components/modals/Send.tsx
Tbaut: why?

packages/ui/src/components/modals/Send.tsx
Tbaut: This is no a string, this is a `EasyTransferTitle`. You can't put any string in there without a type error.

Tbaut:
> Can we at least add input placeholders to give more information about the data we ask?

Sure that's a good idea. Just to be clear, Multix, and multisigs in general are not for ppl new to Polkadot, but it doesn't mean that I can use this as an argument for everything! The goal of this component is to make it so that ppl don't have to use polkadot.js/apps, they had no other option this far. Pjs has an easy flow for setting an identity that looks exactly the same.

image

Tbaut:
I hope you had the chance to test it out, because on Rococo it's now filtered :sob: because of https://github.com/paritytech/polkadot-sdk/pull/1476
It's still valid for all the other chains though

packages/ui/src/components/modals/Send.tsx
Lykhoyda: Yeah, I saw the type, it's still confusing to read. I could understand if you created a type with one-word strings: like active | disabled | focus and used them this way. But for a 2-3 word string, it's probably the case for an enum where you don't use directly the value but the link to it. You can use the whole sentence as a type, but it doesn't mean you should do it. It works both ways, but in my opinion with an enum, it gives better readability and expectation that this is a predefined value.

packages/ui/src/components/EasySetup/IdentitySetIdentity.tsx
Lykhoyda: To rename to identityFields is a good idea, as I was confused with the hook name which has the same name

asnaith:
The recent updates, character limits, placeholder text, and title updates, are nice additions. 

The "Display name" error and highlighted field appear immediately upon loading the modal, which effectively signals that it's a mandatory field. This seems fine to me.

Lykhoyda (approved):

Tbaut:
> The "Display name" error and highlighted field appear immediately upon loading the modal, which effectively signals that it's a mandatory field. This seems fine to me.

Yeah that's not good, I'll change this before merging.

PR #405, opened 2023-10-17
Hotfix for pallet method undefined

Tbaut:
bug in prod. Trying to investigate why, this stops the fire at least.


Tbaut:
I identified where that came from.
I had TX waiting for an approval, set `balances.transfer` and this method has been removed in a runtime upgrade, leading to the app crashing.

PR #410, opened 2023-10-18
Use name registry in recipients list

### LINKED ISSUE #395: Show the named multisig/proxy in the send modal dropdown
**Description**
When making transactions, Multix with provide a drop down with a list of signatories but it will not display other multisigs or proxies. It would be convinient to be able to send to multisigs/proxies from the drop down menu.

See screenshots below
scrend12 47 33_pm_720
screenshot_2023-09-28_at_12 47 27_pm_720

**Steps To Reproduce**

Add clear steps on how to reproduce the bug

1. Go to Multix at 'https://multix.chainsafe.io/'
2. Select the proxy or multisig you wish to send funds from and cick on 'new transaction'
3. When selecting 'send tokens' option you will asked who you want ot send to
4. The list will include other signatories but not multisigs or proxies

Tbaut:
closes #395 

How to test:
In New tx > Send tokens

In the "to" field you'll have the addresses from your name registry too.

### DIFF
--- a/packages/ui/src/hooks/useAccountBaseFromAccountList.tsx
+++ b/packages/ui/src/hooks/useAccountBaseFromAccountList.tsx
@@ -1,15 +1,26 @@
 import { useMemo } from 'react'
 import { AccountBaseInfo } from '../components/select/GenericAccountSelection'
 import { useAccounts } from '../contexts/AccountsContext'
+import { useAccountNames } from '../contexts/AccountNamesContext'
 
-export const useAccountBaseFromAccountList = () => {
+interface Params {
+  withAccountsFromAddressBook: boolean
+}
+
+export const useAccountBaseFromAccountList = (params?: Params) => {
+  const { withAccountsFromAddressBook = false } = params || {}
   const { ownAccountList } = useAccounts()
+  const { accountNames } = useAccountNames()
 
   const accountBase = useMemo((): AccountBaseInfo[] => {
-    return (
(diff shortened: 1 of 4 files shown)


packages/ui/src/hooks/useAccountBaseFromAccountList.tsx
```diff
@@ -1,15 +1,26 @@
...
 import { AccountBaseInfo } from '../components/select/GenericAccountSelection'
 import { useAccounts } from '../contexts/AccountsContext'
+import { useAccountNames } from '../contexts/AccountNamesContext'
 
-export const useAccountBaseFromAccountList = () => {
+interface Params {
+  withAccountsFromAddressBook: boolean
+}
+
+export const useAccountBaseFromAccountList = (params?: Params) => {
```
Lykhoyda, line 10: can we use here object destruction?
```ts
({ withAccountsFromAddressBook = false }: Params) => 
```

Tbaut: This is what I usually prefer too, but in this case, it doesn't allow Param to be undefined. I wanted to keep `Params | undefined` so as not to change anything in the `useAccountBaseFromAccountList` calls that are happening around. And because we can't destructure an undefined object, and afaik there's also no way to say something like `Param | undefined` and at the same time set `{ withAccountsFromAddressBook = false }` as default value.. that's why I went for this.

Lykhoyda (approved):

PR #415, opened 2023-10-19
Add tests for setIdentity feature

### LINKED ISSUE #408: Test the `setIdentity` easy flow
We should add a test with an account that to
- connecting with a multisig/pure that doesn't have an identity
  - verify that the `setIdentity` is visible and usable in the header
  - verify that there are no error when clicking on it
  - if filling another field than the "Display" then an error appears and "send" is disabled
  - if we put more than 32 char, then an error appears  and "send" is disabled
  - we can submit the TX and check the payload if possible (we can then reject the TX in the extension because it won't go through anyways)

- connecting with a multisig/pure that has already an identity, we get specific identity set. already
  - all of the above as well

Tbaut:
closes #408 

As explained in the test, our testing is limited by the fact that Rococo now filters the `identity` calls. I can still test with a multisig that already had an identity before, and check that the outgoing call data is the one I expect.


packages/ui/cypress/tests/setIdentity.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const initAndConnect = (account: InjectedAccountWitMnemonic) => {
+  cy.visit(landingPageUrl)
+  cy.initExtension([account])
+  clickOnConnect()
+  cy.connectAccounts([account.address])
+}
+
+describe('Set an identity', () => {
+  it('Can set an identity from the dropdown', () => {
```
asnaith, line 19:
```suggestion
  it('Can set an identity from the options menu', () => {
```

packages/ui/cypress/tests/setIdentity.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+    sendTxModal.setIdentitySection().should('be.visible')
+
+    // every field is empty to start with
+    sendTxModal.setIdentityFieldInput('display').should('have.value', '')
+    sendTxModal.setIdentityFieldInput('legal').should('have.value', '')
+
+    // typing into legal with empty display should show an error
+    sendTxModal.setIdentityField('legal').type('le')
+    sendTxModal.sendTxError().should('contain', 'Display name is required')
+    sendTxModal.setIdentityFieldElement('display', 'label').should('havInpute.class', 'Mui-error')
```
asnaith, line 34: My approach to this would have been to put a locator on the error itself then in test yield the parent element of the applicable field input, use cy.within() to narrow the scope, and then search for the error element inside there, you probably wouldn't need setIdentityField then. 

This is all very readable though so happy to keep it like it is.

packages/ui/cypress/tests/setIdentity.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+    sendTxModal.setIdentityFieldElement('display', 'div').should('havInpute.class', 'Mui-error')
```
asnaith, line 35: My approach to this would have been to put a locator on the error itself then in test yield the parent element of the applicable field input, use cy.within() to narrow the scope, and then search for the error element inside there, you probably wouldn't need setIdentityField then.

This is all very readable though so happy to keep it like it is.

packages/ui/cypress/support/page-objects/sendTxModal.ts
```diff
@@ -1,6 +1,14 @@
...
-  fieldTo: () => cy.get('[data-cy=field-to]'),
-  fieldAmount: () => cy.get('[data-cy=field-amount]'),
-  buttonSend: () => cy.get('[data-cy=button-send]')
+  sendTokensFieldTo: () => cy.get('[data-cy=send-tokens-field-to]'),
+  sendTokensFieldAmount: () => cy.get('[data-cy=send-tokens-field-amount]'),
+  buttonSend: () => cy.get('[data-cy=button-send]'),
+  setIdentitySection: () => cy.get('[data-cy=section-set-identity]'),
+  setIdentityField: (field: string) => cy.get(`[data-cy=set-identity-field-${field}]`),
+  setIdentityFieldInput: (field: string) => cy.get(`[data-cy=input-set-identity-field-${field}]`),
+  setIdentityFieldElement: (field: string, element = 'label') =>
```
asnaith, line 9:
```suggestion
  setIdentityFieldLabel: (field: string, element = 'label') =>
```

packages/ui/cypress/tests/setIdentity.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+      sendTxModal.buttonSend().should('be.disabled')
+      // expected https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frococo-rpc.polkadot.io#/extrinsics/decode/0x1f0102000412ad770206045069201514711dc2456908b0af226442d475d12a5334e9c4513e001901000564696973056c6565670000000000000000
+      cy.wrap(txRequests[0].payload.method).should(
+        'eq',
+        '0x1f0102000412ad770206045069201514711dc2456908b0af226442d475d12a5334e9c4513e001901000564696973056c6565670000000000000000'
+      )
+    })
+  })
+
+  it.only('Can edit an identity from the new tx button', () => {
```
asnaith, line 72:
```suggestion
  it('Can edit an identity from the new tx button', () => {
```

packages/ui/cypress/tests/setIdentity.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+    const multisigSignatoryWithoutIdentity = setIdentitySignatories[0]
+    initAndConnect(multisigSignatoryWithoutIdentity)
+    // select the right multisig (Alice has a lot)
+    const first5DigitsAddress = setIdentityMultisigs['multisig-with-identity'].address.slice(0, 4)
```
asnaith, line 76: Just a thought - we might have to use this a lot, maybe it's declaring on the the test fixtures?

asnaith:
Minor comments but this is great! 🙌

packages/ui/cypress/support/page-objects/sendTxModal.ts
Tbaut: Nope, because is could be 'label', or 'div' for instance

packages/ui/cypress/tests/setIdentity.cy.ts
Tbaut: ah yeah..

packages/ui/cypress/tests/setIdentity.cy.ts
Tbaut: as a helper function eventually, but I think in the fixture it'd be bloating it maybe?

packages/ui/cypress/tests/setIdentity.cy.ts
Tbaut: the `haveInpute` is super weird! It's supposed to be "have.class"

This is not an error, it's the fields (div and label) that turn red/orange when they are marked as `isError=true`. I don't want to bloat the code too much with test specific things. Of course I could have something like if, there's an error, then `data-cy` should have this. But as I said, this is writing specific code in the UI for the tests, which I think is the limit not to cross.

packages/ui/cypress/tests/setIdentity.cy.ts
Tbaut: Thinking about it, I think it's too simple and the code to get it will be just as long as repeating `slice(0, 4)` Because you'd end up with `getFirst5Char(setIdentityMultisigs['multisig-with-identity'].address)` which to me is the same, with an import spared :shrug:

packages/ui/cypress/support/page-objects/sendTxModal.ts
Tbaut: we use both.

packages/ui/cypress/tests/setIdentity.cy.ts
asnaith: Haha fair point! 👍

asnaith (approved):
Nice expansion of test coverage  🙌

PR #419, opened 2023-10-20
More watched accounts tests

asnaith:
Related to #378

Adds new to test to ensure that
- watched accounts can see but not utilize wallet connect components
- watched accounts can see tx's in the transaction list but cannot see the review button

### DIFF
--- a/packages/ui/cypress/tests/watched-accounts.cy.ts
+++ b/packages/ui/cypress/tests/watched-accounts.cy.ts
@@ -193,7 +193,9 @@ describe('Watched Accounts', () => {
       watchMultisigs['multisig-with-pure'].pureAddress,
       watchMultisigs['multisig-with-pure'].name
     )
-    settingsPage.wallectConnectAccordion().should('be.visible').click()
+    settingsPage.wallectConnectAccordion()
+      .should('be.visible')
+      .click()
     settingsPage
       .walletConnectAlert()
       .should('be.visible')
--- a/packages/ui/cypress/tests/watched-accounts.cy.ts
+++ b/packages/ui/cypress/tests/watched-accounts.cy.ts
@@ -193,9 +193,7 @@ describe('Watched Accounts', () => {
       watchMultisigs['multisig-with-pure'].pureAddress,
       watchMultisigs['multisig-with-pure'].name
     )
-    settingsPage.wallectConnectAccordion()
-      .should('be.visible')
-      .click()
+    settingsPage.wallectConnectAccordion().should('be.visible').click()
     settingsPage
       .walletConnectAlert()
       .should('be.visible')
--- a/packages/ui/cypress/tests/watched-accounts.cy.ts
+++ b/packages/ui/cypress/tests/watched-accounts.cy.ts
@@ -186,4 +186,39 @@ describe('Watched Accounts', () => {
     multisigPage.accountHeader().should('be.visible')
     multisigPage.newTransactionButton().should('not.exist')
   })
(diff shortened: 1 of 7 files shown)


packages/ui/src/pages/Settings/Settings.tsx
```diff
@@ -50,7 +57,14 @@ const Settings = () => {
...
         onChange={() => onToggle(ACCORDION_WALLET_CONNECT)}
       >
-        <AccordionSummaryStyled expandIcon={<ExpandMoreIcon size={20} />}>
+        <AccordionSummaryStyled
+          expandIcon={<ExpandMoreIcon size={20} />}
+          data-cy={
+            expanded === ACCORDION_WALLET_CONNECT
+              ? 'button-collapse-wallet-connect'
+              : 'button-expand-wallet-connect'
+          }
```
Tbaut, line 66: This is too much specific code for the tests IMHO, and it's repeated.
it'd be better to have something simple
```suggestion
          data-cy="accordion-title-wallet-connect"
```

and then check for specific attributes. In this case there's a good one with `aria-expanded`

```
  .should('have.attr', 'aria-expanded', true);
```

packages/ui/cypress/tests/watched-accounts.cy.ts
```diff
@@ -186,4 +186,39 @@ describe('Watched Accounts', () => {
     multisigPage.accountHeader().should('be.visible')
     multisigPage.newTransactionButton().should('not.exist')
   })
+
+  it('can not utilize wallet connect when only a watched account', () => {
```
Tbaut, line 190: This is better maybe?
```suggestion
  it('can not utilize wallet connect with only watched accounts', () => {
```

packages/ui/cypress/tests/watched-accounts.cy.ts
```diff
@@ -186,4 +186,39 @@ describe('Watched Accounts', () => {
...
+      .should('be.visible')
+      .should(
+        'contain.text',
+        'Please first select a Pure or Multisig that you are part of, to be able to use it with WalletConnect'
+      )
+    settingsPage.walletConnectKeyInput().should('be.disabled')
+    settingsPage.connectDappButton().should('be.disabled')
+  })
+
+  it('can see but not interact with txs when only a watched account', () => {
```
Tbaut, line 208: Same
```suggestion
  it('can see but not interact with txs with only watched accounts', () => {
```

Tbaut (commented):
Small comments, looks good otherwise :)

packages/ui/src/pages/Settings/Settings.tsx
asnaith: Updated to use the simpler tag 👍

I'll just click in this test and check for the accordion default state in a wallet-connect test rather than here as we need to add one there for default connected / default non-connected (non watch mode). Makes more sense to be in a separate feature test rather than be part of the checks here.

packages/ui/cypress/tests/watched-accounts.cy.ts
asnaith: Updated to "can see but not interact with txs when in watched account mode"

packages/ui/cypress/tests/watched-accounts.cy.ts
asnaith: Updated to "can not utilize wallet connect when in watched account mode"

Tbaut (approved):

PR #420, opened 2023-10-20
Add visit with Watch and Name helper

### LINKED ISSUE #418: Add helper function to visit a page and start with a set of defined watched accounts
Watched accounts are stored in local storage  with a list of private key (network agnostic).
We can define a helper function to add some without using the UI, and before landing on a page. This will help speed up tests once we have at least one test that makes sure the addition using the UI is working well.

It's explained here https://docs.cypress.io/api/commands/getalllocalstorage#Get-all-localStorage

```js
cy.visit('https://example.cypress.io', {
  onBeforeLoad(win) {
    win.localStorage.setItem('key', 'value')
  },
})
```

Tbaut:
closes #418 

- Refactor a bit the tests to help for readability (hopefully)
- use the helper function once we've tested the actual manual way
- in case you didn't know you can get the publicKey of any account with e.g https://kusama.subscan.io/tools/format_transform

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,4 @@ builds
 # cypress
 packages/ui/cypress/screenshots/*
 packages/ui/cypress/videos/*
+packages/ui/cypress/downloads/*
--- a/packages/ui/cypress/fixtures/accounts.ts
+++ b/packages/ui/cypress/fixtures/accounts.ts
@@ -1,3 +1,0 @@
-export const addresses = {
-  Alice: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'
-}
--- a/packages/ui/cypress/fixtures/knownMultisigs.ts
+++ b/packages/ui/cypress/fixtures/knownMultisigs.ts
@@ -1,9 +1,9 @@
-import { injectedAccounts } from './injectedAccounts'
+import { testAccounts } from './testAccounts'
 
 export const knownMultisigs = {
   'test-multisig-1': {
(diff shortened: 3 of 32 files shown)


asnaith:
This is a nice readability improvement to the commands and assertions and the helper is efficient for setting up state rather than performing it all in the UI (when that's not what we are actually testing) 🚀

Initially, I thought it was a little bit strange taking in the URL in the same function that is stubbing the data but knowing how we're going to utilize this further it does make sense.

asnaith:
Thanks for this! Looks good to merge as soon as the merge conflicts are resolved

asnaith (commented):
Thanks for this! Looks good to merge as soon as the merge conflicts are resolved

Tbaut:
> Initially, I thought it was a little bit strange taking in the URL in the same function that is stubbing the data

If you check the function, adding any localstorage item is done right `onBeforeLoad` the `visit` and is actually part of the same function. So there's not much we could have done. Localstorage has to before the page is loaded, otherwise you'd probably have to trigger a refresh which is costly in time (reconnect to the node, accounts, check multisigs).

I'd appreciate an "Approve" so that I can merge.

asnaith (approved):

asnaith:
> I'd have appreciated an "Approve" so that I can merge (doing it the right way here and waiting since there's no emergency).

@Tbaut Oops, didn't mean to draw things out. It was well-intentioned, after spotting merge conflicts from the updated branch that yours was based on.

Going forward, I'll trust that any existing conflicts will be taken care of and approve accordingly if nothing else outstanding

Issue #400, opened 2023-10-13
Indexer leak potential. Try to understand why


Tbaut:
The indexer has had some days with low activity, very high CPU usage on the API.
Those are oct 7 and 8.

Despite restarting the squid, the level remained super high.
![image](https://github.com/ChainSafe/Multix/assets/33178835/e9d02515-8888-431e-b68d-f513eb6b99f1)

Talking to the subsquid team, 
> It seems that from 6 to 8 October it was generating heavy database queries (since DB CPU was also quite high), and then it just stopped doing that

To my question

> > Could it be related to the fact that my queries are subscriptions, and there's a leak somewhere there?

> The answer is yes

And further advice
> Looks like in your case the reason of app behaviour change can be only related to a change in a usage pattern. 
Large number of users, large subscription query responses, etc.
I would suggest to simulate the load your app gets and investigate its behaviour locally.

Not sure if we can analyze the type of queries that were made on these days. Maybe @mpetrunic can help.

Tbaut:
None of what I've done in #424 and #425 actually changed anything, there are still some random spikes, although somewhat more rare it seems. The Subsquid team told in on Telegram that they will allow to track what queries are made, in the server logs. Let's see if/when this is delivered.

Tbaut:
Asking again, they promised to do it in the next days.

The issue could be that there are lots of queries, which could still be done on low trafic. 

Another avenue to check would be slow query logs on postgres. E.G: 

https://www.cybertec-postgresql.com/en/3-ways-to-detect-slow-queries-in-postgresql/

The `--sql-statement-timeout 3000` flag that we have should however help.. But it doesn't hurt to check.

PR #424, opened 2023-10-24
Manual reconnect button

Tbaut:
related to #400 

After exploring many other potential issues this could come from (generally more complex as well), I'm exploring one that could be the automatic refresh that can't be stopped and could actually spam our server.

I'm therefore removing the auto-refresh that I implemented a while ago, and add a better copy + reconnect button that hopefully makes the life of the couple affected persons better than previously. We will see if the spikes of demand on the server ever decrease. It's a shot in the dark unfortunately.

To test and see the error, change in `constants.ts` the `wsGraphqlUrl` of a network to point to v6 (it's offline) instead of v7, and visit this network.

image

### DIFF
--- a/packages/ui/src/contexts/MultiProxyContext.tsx
+++ b/packages/ui/src/contexts/MultiProxyContext.tsx
@@ -42,6 +42,7 @@ export interface IMultisigContext {
   getMultisigByAddress: (address: string) => MultisigAggregated | undefined
   getMultisigAsAccountBaseInfo: () => AccountBaseInfo[]
   selectedIsWatched: boolean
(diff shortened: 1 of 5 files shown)


asnaith (approved):
It's a good theory to test out, that's for sure! 

The button is working well. I couldn't test by changing wsGraphqlUrl because every time I would change it back to the correct url to test the refetch ability the vite reload automatically made the content load).

I found a different way - just turn off wifi before multisigs have loaded, wait for the button to appear, re-enable wifi and click the button 😄

Lykhoyda (approved):
Looks good! 👍

PR #426, opened 2023-10-26
Add wiki link info about reserved funds

### LINKED ISSUE #414: Wiki article regarding reserved funds and link to it in the UI
This is a very common question that comes back over and over. Users are surprised that you need 40 DOTs to submit the creation of a multisig with proxy. We have clear errors in the interface, but the story to explain it all doesn't fit in an error message. We should point to a wiki article explaining that:

> There are reserved funds for anything that is stored on the blockchain, that's valid for many things including setting an identity. This is an anti-spam measure so that the blockchain doesn't get bloated. On Polkadot it's 20 DOTs. A reserved amount it required for
> 1. pure proxy
> 2. multisig transaction

> The 1. is reserved for as long as you have a pure proxy, and until it's killed. needless to say these funds will be reserved for a while. The 2 is for the multisig transaction that you are about to submit. This will be returns as soon as the transaction is approved/rejected by the multisig. So this 2. is for a very short amount of time.

Tbaut:
closes #414 
Wiki article: https://github.com/ChainSafe/Multix/wiki/Why-are-funds-reserved%3F

image
image

---

Submission checklist:

#### Layout

#### Compatibility

No logic changed here


packages/ui/src/utils/getErrorMessageReservedFunds.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <a
+        href={wikiLinkReservedFunds}
+        about="_blank"
+      >
+        More info.
+      </a>
+    </span>
+  ) : (
+    ''
+  )
```
Lykhoyda, line 23: It's probably better to move it as a component for clarity:
```
const ReservedMessage: <{ reservedAmount?: string }> = ({ reservedAmount }) => {
  if (!reservedAmount) return null;

  return (
    <span>
      Note that it includes ${reservedAmount} that will be reserved.{' '}
      <a
        href={wikiLinkReservedFunds}
        target="_blank"
        rel="noopener noreferrer"
      >
        More info.
      </a>
    </span>
  );
};
```

packages/ui/src/utils/getErrorMessageReservedFunds.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <span>
+      The {identifier} doesn't have the required {requiredBalanceString} to submit this transaction.{' '}
+      {reservedMessage}
```
Lykhoyda, line 27: Also, it's a small optimisation to render component here ReservedMessage, otherwise it will recalculate reservedMessage every render

packages/ui/src/utils/getErrorMessageReservedFunds.tsx
```diff
@@ -0,0 +1,30 @@
...
+  reservedString?: string
+) => {
+  if (!requiredBalanceString) return ''
+
+  const reservedMessage = reservedString ? (
+    <span>
+      Note that it includes ${reservedString} that will be reserved.{' '}
+      <a
+        href={wikiLinkReservedFunds}
+        about="_blank"
```
Lykhoyda, line 16: It's also probably intended to be target="_blank", although it's a good Berlin club that was on your mind 😅

Lykhoyda: There is a good practice to add rel="noopener noreferrer" for security reason

Lykhoyda (changes requested):
Small changes requested to the getErrorMessage component

packages/ui/src/utils/getErrorMessageReservedFunds.tsx
Tbaut: totally!

Tbaut:
great suggestions, thanks

Lykhoyda (approved):
Looks good 👍

PR #428, opened 2023-10-30
Link to specific accounts from the address bar

### LINKED ISSUE #266: Be able to link to a multisig/pure proxy
This is a feature request from a real user. They'd like to be able to share a link to a multisig/pure so that other ppl can watch it right away, and follow the process of a transaction.

---
How I see this happen:
- create a new url param in the home with `?account=[string]`. This account could be a multisig or a pure (if anything else we should show an error? See edge cases)
- add some logic in the Home to:
  - check if this account is already watched, if not, add it.
  - select it if it's in the multiproxy list.
- Upon manual multiproxy change using the dropdown selector, we should update the url

Some edge cases I can think about now:
- the account is not a multisig/proxy, meaning that even after adding it to the watch accounts, the multiproxy list doesn't contain it (either as a multisig or as part of a pure) then we should show an error. 
- however, when we add a watch account, it takes time for the UI to fetch the new multisigs list and actually refresh. This could be an issue as we would show an error but then refresh to show the desired account.

---

The above has the best UX. If this doesn't work, here's another idea:
This would be a dedicated link in the 3 dots menu > share
Create a dedicated route `/share/[string]` and have the whole logic in this dedicated page to
- check if it's in the watched/list already, if yes select (with the localStorage `multix.selectedMultiProxy` that already exists) and go to home
- if not, query the DB synchronously, wait for an answer, if valid select and go to Home
- if not valid show an error.

Tbaut:
closes #266 

This was such a tricky change, that I thought it should go with a full suite of tests.
Now because our cypress extension plugin didn't actually support what I needed, I had to make it evolve, and allow users to connect an extension, then refresh the page without the need to connect again.

This made the scope of this PR explode.. but it was only in order to actually push many many tests. So I hope it alleviates the annoyance.

Note that there's some refactoring to do to make the extension plugin easier to use. I'll do this in another PR see https://github.com/ChainSafe/Multix/issues/433

---

Submission checklist:

#### Layout

#### Compatibility


asnaith (approved):
Really fantastic to deliver this with the supporting UI tests! Thank you for capturing all of those different scenarios

Tbaut:
I think one area to improve is the amount of info and explanations when users land on the error page. They may feel lost and the "reset" button may feel weird. Open to suggestions (maybe for a follow-up pr). 

We should probably have a "why do I see this" accordion or something explaining that they landed here with an account selected, that they don't have neither in their Multisig list, nor as watched accounts. And then we can do some check and suggest to connect their accounts if not done or watch the account if not done yet.

packages/ui/src/hooks/useDisplayError.tsx
```diff
@@ -49,6 +51,30 @@ export const useDisplayError = () => {
...
+          <Button
+            data-cy="button-reset-linked-address"
+            onClick={() =>
+              setSearchParams((prev) => {
+                prev.delete('address')
+                return prev
+              })
+            }
+          >
+            Reset
```
Lykhoyda, line 71: I agree that the "Reset" looks weird here, maybe "Back to home" will be more understandable in terms of UI.

packages/ui/src/hooks/useSwitchAddress.tsx
```diff
@@ -0,0 +1,77 @@
+import { useCallback, useEffect, useMemo } from 'react'
+import { useSearchParams } from 'react-router-dom'
+import { useMultiProxy } from '../contexts/MultiProxyContext'
+
+export const useSwtichAddress = () => {
```
Lykhoyda, line 5: typo

Lykhoyda (approved):
Looks good! Great feature!

There is a small suggestion about the button name when the account is not found and typo to fix. The Transaction test in Cypress will not pass due to Roccoco, so needs to be skiped for now (known issue)

packages/ui/src/hooks/useDisplayError.tsx
Tbaut: Yes, I need to create an issue for this. Agreed that Back to Home would already be better.

PR #434, opened 2023-11-09
Fix date displayed with no Tx

### LINKED ISSUE #430: No transaction but a date is displayed
I can't pinpoint a specific case for which this happens, but I've seen it enough to log an issue. We need to investigate why/how this happens.
image

Tbaut:
closes #430 

This was more complex than expected. The problem was related to a race condition between
- getting the pending TX from the chain (querying the multisig storage)
- getting the call data info from the chain (involving querying older blocks and get the calldata from the `asMulti` calls)

The second call can take time, and when changing accounts, a reset wouldn't be performed. Now since Promises can't be stopped, this resulted in the second promise resolving with outdated data. I solved it in making the 2 calls sequential and less dynamic. It's not really an issue since they should strictly follow each other. I've been consolidating all this logic which used to be separated, into the `usePendinTx` hook. It's a lot, but it should be more comprehensible already.

To see the bug, you'd have to:
- have many pending TXs (I had 10) and at least 1 from the multisig (as opposed to from the proxy). You can try by watching `5EMm18Z8WWWT2wit1RxpoZv39goPdYmSksnDTePYAzrUhdJv` on Rococo.
- then you should select the account with the many tx in the UI, wait for them to show, and switch to an account that doesn't have any pending tx. --> you'd see the date with no tx.
---

Submission checklist:

#### Compatibility


asnaith (approved):
Nice! The fix works successfully, I can no longer recreate the issue on this branch

packages/ui/src/hooks/usePendingTx.tsx
```diff
@@ -1,39 +1,169 @@
...
+  timestamp?: Date
+  multiProxyAddress?: string
+}
+
+type AggGroupedByDate = { [index: string]: CallDataInfoFromChain[] }
+
+export const getMultisigInfo = (c: ISanitizedCall): Partial<CallDataInfoFromChain>[] => {
+  const result: Partial<CallDataInfoFromChain>[] = []
+
+  const getCallResult = (c: ISanitizedCall) => {
```
Lykhoyda, line 38: is it some golang practice to call a param with single letter ? :)

packages/ui/src/hooks/usePendingTx.tsx
```diff
@@ -1,39 +1,169 @@
...
+    if (typeof c.method !== 'string' && c.method.pallet === 'multisig') {
```
Lykhoyda, line 39: can we create a method isMultisigCall for example that the code will be self explanatory?

Lykhoyda (approved):
@Tbaut looks good! I pushed the PR where I replaced the old AggregatedData with the new interface CallDataInfoFromChain. As I had errors during the build. Please confirm if it's correct

packages/ui/src/hooks/usePendingTx.tsx
Tbaut: I think this is not specific to a language.. made this more readable in [ee33cfd](https://github.com/ChainSafe/Multix/pull/434/commits/ee33cfd0fa73946114ee94ba7ee4b5a04283526f)

packages/ui/src/hooks/usePendingTx.tsx
Tbaut: This is only used here, and I'm not a fan of creating a specific function for a not reused one-liner.
This would open the door to then use such function everywhere, like at the next line to check whether it's a `multisig.asMulti`.. effectively opening a can of worms.

Tbaut:
Thanks for the review. I'll merge this PR next week as it's touching quite some critical logic, and I'm not very available in the next days if something goes wrong.

PR #436, opened 2023-11-09
Launch Chopsticks forking Kusama and an indexer pointed at it in tests

### LINKED ISSUE #394: Experiment running our own fork of any network in the CI with Chopsticks
We should try to 
- fork Rococo network with [Chopsticks](https://github.com/AcalaNetwork/chopsticks)
- run an indexer against this fork (arrow-squid should do)
- start with a blank page every single test & test the multisig creation flow, and signing a TX

Tbaut:
closes #394 

The CI now has
- a Postgres for the indexer
- Chopsticks funding 4 accounts and forking Kusama
- Subsquid indexer pointing at Chopsticks, and running alongside a graphQL server
- a minimal test with the full creation of a multisig, with proxy, with 3 of the funded accounts

We need to make this test encompass many more scenarii.
Unfortunately, all these things running in parallel means that the output in CI is very messy.
To test locally, I also made sure we have a simple command. It's `npm run start:chopsticks-test-build-and-launch-all` which will (quoting the README)
```bash
# Launch the full setup. It will:
# - delete any previous Chopsticks database
# - launch a docker with the indexer db
# - launch a Chopsticks node
# - build and launch a subsquid indexer pointed at Chopsticks
# - launch a graphQL server for the squid
# - launch the UI with local env pointed at Chopsticks
```


asnaith:
Just checked out the branch to test `npm run start:chopsticks-test-build-and-launch-all`, and it's working great for me. Thanks for this 🙏

packages/ui/cypress/support/page-objects/landingPage.ts
```diff
@@ -13,6 +13,7 @@ export const landingPage = {
   noAccountFoundError: () => cy.get('[data-cy=label-no-account-found]', { timeout: 10000 }),
   linkedAddressNotFound: () => cy.get('[data-cy=label-linked-address-not-found]'),
   resetLinkedAddressButton: () => cy.get('[data-cy=button-reset-linked-address]'),
+  firstMultisigCreationText: () => cy.get('[data-cy=text-first-multisig-creation'),
```
asnaith, line 16: Very minor conventions point but we usually use "label" instead of "text" in the page objects / locators (checking text of label reads better in test than checking the text of text)

packages/ui/cypress/support/page-objects/multisigPage.ts
```diff
@@ -1,5 +1,5 @@
 export const multisigPage = {
-  accountHeader: () => cy.get('[data-cy=header-account]'),
+  accountHeader: (timeout = 4000) => cy.get('[data-cy=header-account]', { timeout }),
```
asnaith, line 2: I think this one can be removed now, it is declaring the default anyway

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+    newMultisigPage.nameInput().type('Some Name')
+    newMultisigPage.nextButton().should('contain', 'Next').click()
+    newMultisigPage.nextButton().should('contain', 'Create').click()
+    cy.getTxRequests().then((req) => {
+      const txRequests = Object.values(req)
+      cy.wrap(txRequests.length).should('eq', 1)
+      cy.wrap(txRequests[0].payload.address).should('eq', address1)
+      newMultisigPage.nextButton().should('not.exist')
+      newMultisigPage.creatingLoader().should('exist')
+      console.log(txRequests)
```
asnaith, line 42: I think this was probably for debugging and can be removed? (If not lets use cy.log?)

packages/ui/src/components/SuccessCreation.tsx
```diff
@@ -9,7 +9,7 @@ interface Props {
 const SuccessCreation = ({ className }: Props) => {
   return (
     <Box className={className}>
-      <h1>Multisig creation in progress...</h1>
+      <h1 data-cy="text-first-multisig-creation">Multisig creation in progress...</h1>
```
asnaith, line 12:
```suggestion
      <h1 data-cy="label-first-multisig-creation">Multisig creation in progress...</h1>
```

packages/ui/cypress/support/page-objects/multisigPage.ts
Tbaut: nah, it's a JS thing, that let you define a param, and if it's undefined, then it'll be 4000. I actually use it in the creation to wait for 10s `multisigPage.accountHeader(10000)`. Let me check if things pass without.

packages/ui/cypress/tests/multisig-creation.cy.ts
Tbaut: yup, stray comment I forgot to remove

asnaith (approved):
I left a few minor comments that's all. 

This is great, the closest we can get to live testing and with no mocks. 🙌

Thank you for all the effort with this, it's well worth it 🚀

.github/workflows/cypress-tests.yaml
```diff
@@ -25,14 +32,35 @@ jobs:
       - name: Install and show git version
         run: apt-get -y install git && git --version
```
mpetrunic, line 33: why do you need git installation? Isn't it already present on all github runners?

mpetrunic (approved):

.github/workflows/cypress-tests.yaml
Tbaut: Good call, I didn't question the status quo, it's old and probably not needed at all. Let me see what happens if removed.

Tbaut: it wasn't needed :pray:

.github/workflows/cypress-tests.yaml
```diff
@@ -3,11 +3,18 @@ on:
...
   cypress-run:
     runs-on: ubuntu-latest
     container:
       image: cypress/browsers:latest
+    services:
+      postgres:
+        image: postgres:14
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
```
Lykhoyda, line 17: Should we also move it to secrets?

packages/ui/cypress/support/Extension.ts
```diff
@@ -24,7 +24,7 @@ export type TxRequests = Record<number, TxRequest>
 export type AuthRequests = Record<number, AuthRequest>
 
 export type EnableRequest = number
-
+const ss58Format = 2
```
Lykhoyda, line 27: It probably should be KUSAMA_S58_FORMAT

Tbaut: Agreed. Actually I removed it since it's not needed.

Lykhoyda (approved):
Thank you for such a great improvement!

PR #437, opened 2023-11-14
Support Joystream

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/438

---

Submission checklist:

#### Layout

#### Compatibility


asnaith (changes requested):
Looking good for the most part, I was able to create a multisig (without proxy as it says proxies are not supported on Joystream) and send a tx. 

However, there is an app crash if you try to set identity either from the options menu on the main page or the set identity dropdown on the "Send tx" modal. 

CleanShot 2023-11-15 at 15 27 48@2x

Tbaut:
Great find. I guess they don't have the identity pallet as well, and Multix assumes it's there. I'll fix, and add a test.

Tbaut:
Aaaand finally fixed the tests!

asnaith (approved):
Looks great, can no longer see the identity options...but that's not new information because your new test is already telling you that, awesome! 😎

🚀

packages/ui/src/utils/getAsMultiTx.ts
```diff
@@ -0,0 +1,28 @@
...
+interface Params {
+  api: ApiPromise
+  threshold: number
+  otherSignatories: string[]
+  tx?: SubmittableExtrinsic<'promise', ISubmittableResult> | HexString
+  weight?: Weight
+  when?: MultisigStorageInfo['when']
+}
+export const getAsMultiTx = ({ api, threshold, otherSignatories, tx, weight, when }: Params) => {
+  return api.tx.multisig.asMulti.meta.args.length === 6
```
Lykhoyda, line 16: Can we make a "6" a constant that will describe itself? as it looks as magic number now :)

Lykhoyda (approved):
looks good! Please use some constant in getAsMultiTx file

packages/ui/src/utils/getAsMultiTx.ts
Tbaut: Good point, it is indeed a magic number as it's actually not future proof. Unfortunately we have no other choice. Joystream will migrate in the next months, and we can eventually remove this "if". I'll make this a constant such as `PARAM_LENGTH_OLD_ASMULTI`

PR #440, opened 2023-11-23
chore: add network switch test

### LINKED ISSUE #417: Test network switch
Check that we can switch networks. This can all be done with watch accounts
- the url changes
- the multiproxy selector disappears and a loader is there
- we see the "connection to node..."
- the watch accounts change the address format

Lykhoyda:
closes #417


packages/ui/cypress/support/commands.ts
```diff
@@ -109,6 +110,16 @@ Cypress.Commands.add('connectAccounts', (accountAddresses = [Account1] as string
   })
 })
 
+Cypress.Commands.add('addWatchAccount', (address: string, name?: string) => {
```
Tbaut, line 113: This should not be used all over because it's slow, manually adding a watch account is already tested extensively. Please use `visitWithLocalStorage` to use watched accounts, if what you want is test another functionality.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+const { name: testAccountName, address: testAccountAddress } =
+  testAccounts['Multisig Member Account 1']
+
+describe('Network can be switched', () => {
+  it('should switch account using menu', () => {
+    cy.visit(landingPageUrl)
+
+    landingPage.watchAccountButton().click()
+
+    cy.url().should('contain', 'settings?network=rococo')
```
Tbaut, line 17: I'm not sure how pedantic we should be.
having `settings?address=xxx&network=rococo` would fail but be valid. Should we just check for `network=rococo`?

```suggestion
    cy.url().should('contain', 'network=rococo')
```

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+import { topMenuItems } from '../support/page-objects/topMenuItems'
+
+const { name: testAccountName, address: testAccountAddress } =
+  testAccounts['Multisig Member Account 1']
+
+describe('Network can be switched', () => {
+  it('should switch account using menu', () => {
+    cy.visit(landingPageUrl)
+
+    landingPage.watchAccountButton().click()
```
Tbaut, line 15: please use `visitWithLocalStorage` to make it optimized instead.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+      accountDisplay.addressLabel().contains(testAccountAddress.slice(0, 5))
+    })
+
+    topMenuItems.desktopMenu().within(() => topMenuItems.networkSelector().click())
+    topMenuItems.networkSelectorOption('kusama').click()
+    topMenuItems
+      .desktopMenu()
+      .within(() => cy.waitUntil(() => topMenuItems.multiproxyLoader().should('not.be.visible')))
+
+    cy.url().should('contain', 'settings?network=kusama')
```
Tbaut, line 32: Same here
```suggestion
    cy.url().should('contain', 'network=kusama')
```

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    settingsPage.accountContainer().within(() => {
+      accountDisplay.identicon().should('be.visible')
+      accountDisplay.nameLabel().should('contain', testAccountName)
+      accountDisplay.addressLabel().contains('Hbiow')
```
Tbaut, line 37: This is hardcoded and not clear to the reader. Either we should re-encode the address with the right ss58 prefix, or probably much simpler and easier to read, have a `kusamaAddress` as part of this test account, and then do a `kusamaAddress.slice(0,5)`?

Tbaut (changes requested):

Lykhoyda:
@Tbaut thank you for the review and suggestion! Please take a look on updated version

Tbaut:
Great!
Let's do more a couple more checks. The issue mentions (maybe not clearly): Let's make sure there's a Multisig in the `MultiProxy` dropdown for rococo (checking and it's there). We can also check that the url contains a `adress=[the expected multisig]`. And then make sure on Kusama there's no multiproxy selector, and no `address=` in the url bar. Check the landing page message (with buttons to create/watch)

Maybe for another issue, but I'm thinking, we should check that 
- add a watched account that has a multisig on Rococo, and add an account that has a multisig on Kusama 
- a certain multiproxy is selected automatically on Rococo (this we kind of see it already, in the current test)
- go on the home page
- Switch network
- verify that the other multiproxy is selected by default.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,44 @@
...
+  address: testAccountAddress,
+  kusamaAddress
+} = testAccounts['Multisig Member Account 1']
+
+describe('Network can be switched', () => {
+  it('should switch account using menu', () => {
+    cy.visitWithLocalStorage({
+      url: settingsPageWatchAccountUrl,
+      accountNames: { [testAccountAddress]: testAccountName as string },
+      watchedAccounts: [testAccountAddress]
```
Tbaut, line 18: Please use the public key here instead of the address, as done in all other tests. The fact that it works with an address is actually a legacy property that could be removed. The UI stores the watched accounts in a way that it's not tied to the network.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,44 @@
...
+      .desktopMenu()
+      .within(() => cy.waitUntil(() => topMenuItems.multiproxyLoader().should('not.be.visible')))
+
+    cy.url().should('contain', 'network=kusama')
+
+    settingsPage.accountContainer().within(() => {
+      accountDisplay.identicon().should('be.visible')
+      accountDisplay.nameLabel().should('contain', testAccountName)
+
+      if (kusamaAddress) accountDisplay.addressLabel().contains(kusamaAddress.slice(0, 5))
```
asnaith, line 41: I don't think this needs to be conditional. We know `kusamaAddress` exists in the test fixture and if the test fails due to the absence of it then it'll be clearer that there's an issue with the data setup. This would be more informative than silently skipping an assertion imo.

Tbaut: oh somehow I didn't see this, agreed.

Lykhoyda:
> Great! Let's do more a couple more checks. The issue mentions (maybe not clearly): Let's make sure there's a Multisig in the `MultiProxy` dropdown for rococo (checking and it's there). We can also check that the url contains a `adress=[the expected multisig]`. And then make sure on Kusama there's no multiproxy selector, and no `address=` in the url bar. Check the landing page message (with buttons to create/watch)
> 
> Maybe for another issue, but I'm thinking, we should check that
> 
> * add a watched account that has a multisig on Rococo, and add an account that has a multisig on Kusama
> * a certain multiproxy is selected automatically on Rococo (this we kind of see it already, in the current test)
> * go on the home page
> * Switch network
> * verify that the other multiproxy is selected by default.

I updated the code:
- check for multisig in dropdown
- Address verification on rococo
- Kusama contains no address in the URL
- Kusama has no multisig selector
- Kusama has a connect button

About the other suggestions, I believe that the test should test a specific functionality like here we are testing the network switch (it's more like an integration test) or complete user flow. Switching the pages back and forth doesn't bring an intuitive understanding of the purpose of the tested functionality. So let's be precise about the topic and scope of the test. 

The scenario you shared can be integrated as a separate test as you want to make sure that the multiproxy selector is working as expected during the network switch. And in my current test case I focus on the network switch behaviour

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const {
+  publicKey: multisigPublicKey,
+  address: multisigAddress,
+  pureAddress: multisigPureAddress,
+  name: multisigName
+} = watchMultisigs['multisig-with-pure']
+
+describe('Network can be switched', () => {
+  it('should switch account using menu', () => {
```
Tbaut, line 16: I'm not sure I understand the purpose of this title here in the "network switch" test. Testing the account switch is something that is either already done elsewhere, or should be done inside an account related test. It seems out of place.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+    cy.visitWithLocalStorage({
+      url: settingsPageWatchAccountUrl,
+      accountNames: { [multisigPublicKey]: multisigName as string },
+      watchedAccounts: [multisigPublicKey]
+    })
+
+    cy.url().should('contain', `network=rococo&address=${multisigPureAddress}`)
```
Tbaut, line 23: We don't want to impose the order of things here. Let's do 2 separate checks
```suggestion
    cy.url().should('contain', 'network=rococo')
    cy.url().should('contain', `address=${multisigPureAddress}`)
```

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    settingsPage.accountContainer().within(() => {
+      accountDisplay.identicon().should('be.visible')
+      accountDisplay.nameLabel().should('contain', multisigName)
+      accountDisplay.addressLabel().contains(multisigAddress.slice(0, 5))
+    })
+
+    topMenuItems.desktopMenu().within(() =>
+      topMenuItems
+        .multiproxySelector()
```
Tbaut, line 33: We can check that the multiproxySelector exists, but I don't think changing the account brings anything to this test.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+    topMenuItems.networkSelectorOption('kusama').click()
+    topMenuItems
+      .desktopMenu()
+      .within(() => cy.waitUntil(() => topMenuItems.multiproxyLoader().should('not.be.visible')))
+
+    cy.url().should('contain', 'network=kusama')
+    cy.url().should('not.contain', 'address=')
+
+    topMenuItems.desktopMenu().within(() => topMenuItems.multiproxySelector().should('not.exist'))
+    topMenuItems.desktopMenu().within(() => topMenuItems.connectButton().should('exist'))
```
Tbaut, line 49: This is also not related to the network selector. the connect button is always there. I'd remove this part.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+    settingsPage.accountContainer().within(() => {
+      accountDisplay.identicon().should('be.visible')
+      accountDisplay.nameLabel().should('contain', multisigName)
+      accountDisplay
+        .addressLabel()
+        .contains(convertRococoAddressToKusama(multisigAddress).slice(0, 5))
```
Tbaut, line 55: iirc, you can use `const kusamaAddress = encodeAddress(multisigAddress, kusamaSS59Prefix)` directly, so that we can don't need an entirely dedicated, and non generic function for this. Address encoding is something very common in all the Polkadot related applications.

packages/ui/cypress/tests/network-switch.cy.ts
Lykhoyda: I think it's a mistake, it should be should switch network

packages/ui/cypress/tests/network-switch.cy.ts
Lykhoyda: For me both ways are fine

packages/ui/cypress/tests/network-switch.cy.ts
Lykhoyda: " Let's make sure there's a Multisig in the MultiProxy dropdown for rococo (checking and it's there)"

If I understood you correctly I'm checking for a multisig selector if the watched multisig exists in the dropdown. Or did you mean to check if a multisig dropdown exists?

Tbaut: Yes, I mean the drop-down exist. I'm juts trying to come up with tests that change between one network and another. In this case, the selected account has a multisig on Rococo, none on Kusama, that's what needs to be checked.

packages/ui/cypress/tests/network-switch.cy.ts
```diff
@@ -0,0 +1,53 @@
+import { accountDisplay } from '../support/page-objects/components/accountDisplay'
+import { settingsPageWatchAccountUrl } from '../fixtures/landingData'
+import { settingsPage } from '../support/page-objects/settingsPage'
+import { topMenuItems } from '../support/page-objects/topMenuItems'
+import { watchMultisigs } from '../fixtures/watchAccounts/watchMultisigs'
+import { encodeAddress } from '@polkadot/keyring'
```
Tbaut, line 6: Let's import it from `@polkadot/util-crypto` as we do elsewhere, which is a direct dependancy, unlike keyring

```suggestion
import { encodeAddress } from '@polkadot/util-crypto'
```

Tbaut (approved):
Nice, there's just a little nit left, looks good otherwise

PR #442, opened 2023-11-26
Use Rhala on rococo

### LINKED ISSUE #441: Switch Rhala to the one on Rococo
Rhala should be indexed using wss://rhala-node.phala.network/ws

Tbaut:
closes #441 
The phala team is now using a new endpoint, after the previous testnet was bricked.
This is already deployed and can be tested live.


Lykhoyda (approved):
Looks good!

PR #444, opened 2023-11-28
fix: menu item wrong displaying in Safari

### LINKED ISSUE #443: Solo chain menu options not displayed correctly (WebKit)
**Description**
Currently the option for "Joystream" in the responsive layout is not displaying correctly, this is so far only reproducible within the Nova App on iOS and Safari on iOS and Safari on macOS in repsonsive layout (see screenshots / video) so seems like a WebKit issue.  

**Steps To Reproduce**

1. Navigate to Multix within Nova app on iOS or within Safari
2. Open the "hamburger" menu to see options
3. Scroll to the "Joystream" option

**Expected Behaviour**
Menu option should be clearly displayed like the others. 

**Screenshots / Video**

https://github.com/ChainSafe/Multix/assets/3982665/01997c69-c0dd-4761-8313-bfe4d4ef0bb2

menu option in safari

menu option in desktop safari

**Test Host Specifics**

- Safari or Nova Wallet mobile app on iOS 17.1.1 
- Safari on macOS using in responsive layout 

**Additional context**
- Cannot reproduce the issue using the web dev tools and responsive layout in other browsers
- Works as intended on Android too

Lykhoyda:
closes #443 

As a Joystream element is alone in the list, It weirdly splits the element into two columns in Safari. Added the fix for one network in the list.

### DIFF
--- a/packages/ui/src/components/select/NetworkSelection.tsx
+++ b/packages/ui/src/components/select/NetworkSelection.tsx
@@ -34,22 +34,26 @@ const NetworkSelection = () => {
   )
 
   const renderNetworks = useCallback(
-    (allowedNetworks: string[]) =>
-      networksToShow
-        .filter(([networkName]) => allowedNetworks.includes(networkName))
-        .map(([networkName, { logo }]) => (
-          <MenuItemStyled
-            key={networkName}
-            value={networkName}
-            data-cy={`select-network-option-${networkName}`}
-          >
-            <ImgStyled
-              alt={`network-logo-${networkName}`}
-              src={logo}
-            />
-            <ItemNameStyled>{networkName}</ItemNameStyled>
-          </MenuItemStyled>
-        )),
+    (allowedNetworks: string[]) => {
+      const filteredNetworks = networksToShow.filter(([networkName]) =>
+        allowedNetworks.includes(networkName)
+      )
+
+      return filteredNetworks.map(([networkName, { logo }]) => (
+        <MenuItemStyled
+          key={networkName}
+          value={networkName}
+          data-cy={`select-network-option-${networkName}`}
+          isOneLine={filteredNetworks.length === 1}
+        >
+          <ImgStyled
+            alt={`network-logo-${networkName}`}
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/select/NetworkSelection.tsx
```diff
@@ -34,22 +34,26 @@ const NetworkSelection = () => {
...
-              alt={`network-logo-${networkName}`}
-              src={logo}
-            />
-            <ItemNameStyled>{networkName}</ItemNameStyled>
-          </MenuItemStyled>
-        )),
+    (allowedNetworks: string[]) => {
+      const filteredNetworks = networksToShow.filter(([networkName]) =>
+        allowedNetworks.includes(networkName)
+      )
```
Tbaut, line 40: My opinion, but I don't think it helps in understanding anything. 
Having `networksToShow.fiter.map` is the same but shorter than `const fileteredNetwowk... and then filteredNetwork.map ` :shrug: 

Maybe that's a naming issue with "filteredNetworks", "networksToShow", "allowedNetworks" not being super meaningful. In any case this change adds more lines to no help in the comprehension IMHO.

Tbaut (commented):
Thanks, left a small comment on the code, and I'll let @asnaith check whether this is fixing the issue for him.

packages/ui/src/components/select/NetworkSelection.tsx
Lykhoyda: I split it as I need to know the amount of networks to apply CSS styles:
```js
isOneLine={filteredNetworks.length === 1}
```

Tbaut: ah my bad. Then let's rename to `displayedNetworks` maybe?

Tbaut (approved):
Good to go code-wise, with just a renaming suggestion

asnaith (approved):
Thanks @Lykhoyda. The fix is working great within Nova app browser, iOS Safari and macOS Safari

PR #448, opened 2023-11-30
Add ability to init the wallet with a predefined allowed origin 

### LINKED ISSUE #433: Refactor the Cypress wallet plugin
We should probably allow to init with enabled origin, so that we don't have to click on connect. (in our case we'll be able to add a localstorage entry to be able to

Tbaut:
closes #433

This allows to reduce a lot of the clicks to connect etc if not needed. Users once they've allowed a website to access their accounts will not have to authorize the connection every single time. This is what this feature enables.

The `visitWithLocalStorage` function which was used all over, now must have a new name. Please suggest anything better if you think `visitCustom` isn't great. The reason why I had to bloat it, is because the extension and the accounts need to be loaded in the window object before the page loads, which is also when we set the custom localStorage elements :man_shrugging:

### DIFF
--- a/packages/ui/cypress/support/Extension.ts
+++ b/packages/ui/cypress/support/Extension.ts
@@ -39,7 +39,7 @@ export class Extension {
     this.allowedOrigins = {}
   }
 
-  init = async (accounts: InjectedAccountWitMnemonic[]) => {
+  init = async (accounts: InjectedAccountWitMnemonic[], allowedOrigin?: string) => {
     this.reset()
     this.accounts = accounts
     await cryptoWaitReady()
@@ -48,6 +48,12 @@ export class Extension {
       // we only add to the keyring the accounts with a known mnemonic
       !!mnemonic && this.keyring?.addFromUri(mnemonic)
     })
+
+    const accountAddresses = accounts.map(({ address }) => address)
+    // if passed along all the accounts will be allowed for this origin
+    if (allowedOrigin) {
+      this.allowedOrigins[allowedOrigin] = accountAddresses
+    }
   }
 
   getInjectedEnable = () => {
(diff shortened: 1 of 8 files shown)


asnaith:
> The `visitWithLocalStorage` function which was used all over, now must have a new name. Please suggest anything better if you think `visitCustom` isn't great. 

As the setup is done before the visit perhaps we could name it this way around?

- setupAndVisit
- configureAndVisit
- injectConfigAndVisit (I like this one but it's long 😅)

asnaith (approved):

Tbaut:
That failing test is annoying. I can reproduce locally, and it sounds like it's actually a bug, but it's not happening in prod,

Lykhoyda:
> > The `visitWithLocalStorage` function which was used all over, now must have a new name. Please suggest anything better if you think `visitCustom` isn't great.
> 
> As the setup is done before the visit perhaps we could name it this way around?
> 
> * setupAndVisit
> * configureAndVisit
> * injectConfigAndVisit (I like this one but it's long 😅)

Agree, we need to come up with a better name for visitCustom method. Maybe `visitWithConfig` or something else

asnaith:
setupAndVisit it is :) 

Looks good to go

Tbaut:
ah I wrote a comment the other day, at least I thought, but didn't click. I went for `setupAndVisit` because "inject" felt weird to me, and keeping it generic lets us add many other things in the future, hopefully without renaming.

now weirdly the test that used to fail is passing... :roll_eyes: if it proves annoying in the future I'll just take another look at it, but it sounded a little like an annoying edge case specific to the test.

Tbaut:
I just had to name it for it to fail. I'll take a look then, before merging.

Lykhoyda (approved):

PR #449, opened 2023-12-01
fix: menu jump during multisig switch

### LINKED ISSUE #387: Jumping/flashing issues in menu bar
From https://github.com/ChainSafe/Multix/pull/385#discussion_r1348566242
When switching between networks, the loader makes the UI jump on desktop and on mobile. While it is important to show users that something is happening as a consequence of their actions, the UI should ideally not jump.

https://github.com/ChainSafe/Multix/assets/33178835/2863d32d-b0e4-4475-b701-e150ea7967b3

Lykhoyda:
closes #387

- Fixes UI jump during multisig switch 
- The loader is not shown anymore for the multiproxy selector
- Overview page has loading status during multisigs fetching

---

Submission checklist:

#### Layout

- [X] Change inspected in the desktop webUIi
- [X] Change inspected in the mobile webUIi

#### Compatibility

- [X] Functionality of change validated with a connected account with multisig
- [ ] Applicable elements hidden/disabled for watched multisigs / pure
- [X] Looks good for solo multisig
- [X] Looks good for multisig with proxy

### DIFF
--- a/packages/ui/src/components/LoadingBox.tsx
+++ b/packages/ui/src/components/LoadingBox.tsx
@@ -3,11 +3,11 @@ import { Box, CircularProgress } from '@mui/material'
 
 interface Props {
   message: string
-  dataCy?: string
+  testId?: string
 }
 
-const LoadingBox = ({ message, dataCy }: Props) => (
-  <MessageBox data-cy={dataCy}>
+const LoadingBox = ({ message, testId }: Props) => (
+  <MessageBox data-cy={`loader-${testId}`}>
     <CircularProgress />
     <div>{message}</div>
   </MessageBox>
--- a/packages/ui/src/components/LoadingBox.tsx
+++ b/packages/ui/src/components/LoadingBox.tsx
@@ -1,0 +1,25 @@
+import { styled } from '@mui/material/styles'
+import { Box, CircularProgress } from '@mui/material'
+
+interface Props {
+  message: string
+  dataCy?: string
+}
+
+const LoadingBox = ({ message, dataCy }: Props) => (
+  <MessageBox data-cy={dataCy}>
+    <CircularProgress />
+    <div>{message}</div>
+  </MessageBox>
+)
+
+export const MessageBox = styled(Box)`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  width: 100%;
+  padding: 1rem;
+`
+
+export default LoadingBox
--- a/packages/ui/src/contexts/MultiProxyContext.tsx
+++ b/packages/ui/src/contexts/MultiProxyContext.tsx
@@ -36,7 +36,7 @@ export interface IMultisigContext {
   selectedMultiProxy?: MultiProxy
   multiProxyList: MultiProxy[]
   isLoading: boolean
-  isFetching: boolean
+  isFetchingMultisigs: boolean
   selectMultiProxy: (multi: MultiProxy | string) => boolean
   selectedHasProxy: boolean
   error: Error | null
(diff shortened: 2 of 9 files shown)


packages/ui/src/hooks/useMultisigsBySignatoriesOrWatchedSubscription.tsx
```diff
@@ -106,5 +106,11 @@ export const useMultisigsBySignatoriesOrWatchedSubscription = ({
   // console.log('subscription data', data)
   //   return <div>Data: {JSON.stringify(data?.multisigCalls)}</div>;
 
-  return { data, isLoading: hasSomethingToQuery && isLoading, error, refetch }
+  return {
+    data,
+    isLoading: hasSomethingToQuery && isLoading,
+    isFetchingMultisigs: isFetching,
```
Tbaut, line 112: I was wondering why not using `isLoading`? From [the docs for isLoading](https://tanstack.com/query/v5/docs/react/reference/useQuery):
>  Is true whenever the first fetch for a query is in-flight
    Is the same as isFetching && isPending

so I think it's not needed to add is, we can use `isLoading`.

packages/ui/src/components/LoadingBox.tsx
```diff
@@ -0,0 +1,25 @@
+import { styled } from '@mui/material/styles'
```
Tbaut, line 1: Nice refactor :pray:

packages/ui/src/contexts/MultiProxyContext.tsx
```diff
@@ -225,6 +226,7 @@ const MultiProxyContextProvider = ({ children }: MultisigContextProps) => {
   const watchedAddressesIds = useAccountId(watchedAddresses)
   const {
     isLoading: isMultisigsubLoading,
+    isFetchingMultisigs,
```
Tbaut, line 229: Beside my previous comment regarding `isLoading`, this is not the only place where things could be loading.
I'll admit this context is way too complex. The reason is query optimisations that obliged us to have actually 2 subscriptions. 1 for the multisigs + proxy, and 1 for the watched pure proxies. 

I think we should use the `isLoading` status and remove this one, but just FYI the other subscription in this file `usePureByIdsSubscription` is also mutating the `multiProxyList` and this is why relying only on this subscription wouldn't work well in some scenario.

Tbaut: Did the behavior feel weird/inconsistent when using isLoading? The isLoading state that we export outside of the context is much more than just a multisig subscription. Also f there is a new account or watched account.re-do the subscription (it's part of the id of the hook) and a whole refetch is happening. If nothing is changing, then because it's a subscription, there's actually no refetching.

Lykhoyda: Changed back to "isLoading" state

packages/ui/src/hooks/useDisplayLoader.tsx
```diff
@@ -15,48 +14,39 @@ export const useDisplayLoader = () => {
 
   if (!isWatchAddressInitialized || !isLocalStorageSetupDone) {
     return (
-      <LoaderBoxStyled data-cy="loader-initialization">
-        <CircularProgress />
-        Initialization...
-      </LoaderBoxStyled>
+      <LoadingBox
+        message="Initialization..."
+        dataCy="loader-initialization"
```
Tbaut, line 19: and these will then change too
```suggestion
        testId="initialization"
```

asnaith (approved):
Just took a look at it, I was awaiting the latest code changes after Thibaut's review before testing it. 

It looks good to merge after the testID convention updates :) 

This is a nice improvement, especially for mobile view

packages/ui/src/hooks/useDisplayLoader.tsx
Lykhoyda: Thank you for your suggestions! Now it looks even better!

Tbaut (approved):
Amazing, thanks!

PR #451, opened 2023-12-04
fix: Align the delete icon for change mutisig modal

### LINKED ISSUE #194: Align the delete icons on "Change Multisig" modal
As described. It would be a visual improvement if all the delete/trash icons were in alignment 

Screenshot 2023-06-06 at 4 40 31 PM

Lykhoyda:
closes #194 
- Aligned delete icon
- Adjust Hover state for the delete button
- Refactored to styled component

---

Submission checklist:

#### Layout

- [X] Change inspected in the desktop web ui
- [X] Change inspected in the mobile web ui

#### Compatibility

- [X] Functionality of change validated with a connected account with multisig
- [X] Looks good for multisig with proxy

### DIFF
--- a/packages/ui/src/components/modals/ChangeMultisig.tsx
+++ b/packages/ui/src/components/modals/ChangeMultisig.tsx
@@ -437,7 +437,7 @@ const ChangeMultisig = ({ onClose, className }: Props) => {
               spacing={2}
             >
               <Box className="loader">
-                {callError ? <ErrorOutlineIcon size={64} /> : <CircularProgress size={64} />}
+                {callError ? <ErrorOutlineIcon size={48} /> : <CircularProgress size={48} />}
                 {!!callError && (
                   <div className="callErrorMessage">
                     {callError.includes('multisig.NoTimepoint')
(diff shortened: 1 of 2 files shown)


Tbaut:
This looks much better!
I just checked on smaller screens and I think we should reduce the outside margins of the "selected" area because this messes up the design, on iphone 11, 12, 13 etc. 

image

asnaith:
Looking great except for the small device / margin issue which Thibaut already mentioned.

 I'm guessing the screenshot below is also the same / similar issue? Here the last two are not in alignment although they don't break out of the parent (this is the "Pixel 7" layout in dev tools). 

CleanShot 2023-12-04 at 11 10 39@2x

I really like the new hover highlighting the entire row, that's a good visual tweak

Lykhoyda:
@Tbaut and @asnaith thank you for the feedback, good catch with the mobile version! I updated the code please try again, especially the Pixel 7 layout as I don't have such multisig with a lof of accounts

Tbaut (approved):
Looking great now, thank you!

A couple notes for the PR
- please remove the checkboxes in the first post if you think they don't apply, so that we don't see them as a "todo" from the implementer that aren't done yet
- regarding the mobile view, this checkbox is exactly there for the reason that implementers should remember to check some of the smallest recent phones, such as the iphones SE. We of course don't want to support every single mobile phone.

asnaith:
> @Tbaut and @asnaith thank you for the feedback, good catch with the mobile version! I updated the code, so please try again. Especially the Pixel 7 layout as I don't have such a multisig with a lot of accounts

@Lykhoyda Yep, I've re-tested with the latest version of code and all looking great now :)

PR #458, opened 2023-12-07
Add Watr parachain

Tbaut:
The Watr indexer is now deployed. You can watch accounts such as: `2u32qBnDgkCdyR26K9Jax7Yf4FQWSY5xLkmbC9jhprFPUKoN`. It's the only multisig I could find there.


asnaith:
I was able to watch the specified account so I am approving tentatively but a couple of questions about this: 

- Do we know if they support the proxy pallet? Because during creation we still show the checkmark for one, if not supported then we should show the unsupported banner like Moonbeam / Joystream 

- I was not able to find a block scanner (specifically subscan) for WATR so we may potentially have an issue with scanner links in the toast notifications?

asnaith (approved):
I was able to watch the specified account so I am approving tentatively but a couple of questions about this:

- Do we know if they support the proxy pallet? Because during creation we still show the checkmark for one, if not supported then we should show the unsupported banner like Moonbeam / Joystream

- I was not able to find a block scanner (specifically subscan) for WATR so we may potentially have an issue with scanner links in the toast notifications?

asnaith:
> * The proxy pallet is detected programmatically. Moonbeam is somewhat of an exception as they have the proxy pallet, but filter pure proxies, in which case we had to have a specific flag.
> * No link will be displayed if there is no subscan explorer available. This is part of the environment, and is visible in the `explorerNetworkName` set to empty here.

Ok sounds good, looks this is good to go then :)

Lykhoyda (approved):
Looks good!

Issue #407, opened 2023-10-18
Enhance `transactions` test


Tbaut:
The test that we have right now in https://github.com/ChainSafe/Multix/blob/c4d4c865b2c7db22b0092690f409f9cabb07c4be/packages/ui/cypress/tests/transactions.cy.ts#L19 is a PoC. We should enhance it with
- in "Makes a balance transfer with Alice"
Once the TX has been submitted verify that
  - it gets displayed eventually (this takes probably 30s)
  - the decoded data is correct and that we show the right sender and balance (in ROC)
  - the signers is correct (e.g that only Alice signed)
  - the threshold is correct (e.g 1/2)
  - we can click on "review"
  - we can only reject with Alice
  - can approve with Bob

PR #462, opened 2023-12-10
Hotfix: Balance display should not show the token and should not format the amount

### LINKED ISSUE #461: The main token with decimal formating is displayed for unrelated asset transfer
Here is an example on polkadot asset hub. The extrinsic assets.transferKeepAlive is not about DOTs, although it's a Balance type.

image

we should remove this functionality for the Asset-hubs and Hydra as the same happens, or figure out when this makes sense or not.

Tbaut:
closes #461 
Only some very specific calls should have the balance displayed nicely with the token and the amount of decimals. Check the changes in `constants.ts` to see which ones. These will be nice in the manual extrinsics and in the call info, while all others should be in plank.
 
You can watch `7J9rSWpjfQjSYr1QDKPr6KjhnC2b2kLWfyBTiTpMgkNvD7vr` on Hydra and see in prod how it shows HDX wrongly, and on this branch it doesn't.


Lykhoyda (approved):

Tbaut:
Ok I added a some test to check that

    - the displayed value of an extrinsic not in the list is in plank
   -  the input value for the same, is also in plank
    - the input value for an extrinsic in this list is in the denominated token.

The check that the denomination is right should probably be made in another test, I didn't have enough balance on hydra to create it 😅 see https://github.com/ChainSafe/Multix/issues/407

asnaith:
Nice fix and tests 👍

Issue #422, opened 2023-10-23
Add UI Tests: Wallet Connect feature


asnaith:
**Scenario:** Validate that wallet connect sessions can be established, rejected, and disconnected in Multix. 

**Acceptance Criteria:**

- New cypress tests are added, suggestions:
  - [x] can see the default wallet connect state when no wallet connected
  - [x] can see the default wallet connect state when wallet is connected
  - [ ] can see the wrong network warning on wc connection modal
  - [ ] can see the correct accounts on wc connection modal
  - [ ] can successfully establish a wallet connect session
  - [ ] can disconnect from a wallet connect session
  - [ ] can see the wallet connect wiki link
  - [ ] can see error modal if typing incorrect wallet connect key

asnaith:
@Tbaut I think we can only connect to the hydra testnet if running locally or we have a multisig on Polkadot. 

Do you have any thoughts on how we can create a test that fully establishes a session? Also, IIRC we'd need a new wc key each time too.

Tbaut:
Even harder than connecting to a hydra testnet, the connection to another Dapp in an other tab is simply not possible. Testing walletconnect is totally out of the table for now I'd say.

Tbaut:
Closing this as there won't be any way to test this in the foreseeable future

asnaith:
> Closing this as there won't be any way to test this in the foreseeable future

Ah yeah, good call. We have some basic accordion / state inspections and will have to leave it there.

PR #464, opened 2023-12-11
Basic tests for wallet connect expected state

asnaith:
related to #422

### DIFF
--- a/packages/ui/cypress/tests/wallet-connect.cy.ts
+++ b/packages/ui/cypress/tests/wallet-connect.cy.ts
@@ -1,0 +1,53 @@
+import { settingsPageUrl } from '../fixtures/landingData'
+import { settingsPage } from '../support/page-objects/settingsPage'
+import { topMenuItems } from '../support/page-objects/topMenuItems'
+import { testAccounts } from '../fixtures/testAccounts'
+
+describe('Wallet Connect', () => {
+  it('can see expected wc state when a wallet is not connected', () => {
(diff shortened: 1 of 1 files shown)


asnaith:
I found these tests in a stash, think I started working on them before realizing we couldn't go much further with wc 

Thought we might as well add these basic checks as they are.

packages/ui/cypress/tests/wallet-connect.cy.ts
```diff
@@ -0,0 +1,53 @@
...
+    settingsPage.connectDappButton().should('be.disabled')
+  })
+
+  it('can see expected wc state when wallet is connected', () => {
+    cy.setupAndVisit({
+      url: settingsPageUrl,
+      extensionConnectionAllowed: true,
+      injectExtensionWithAccounts: [testAccounts['Multisig Member Account 1']]
+    })
+    topMenuItems.settingsButton().click()
```
Tbaut, line 29: That one shouldn't be needed if you're connected already and visited the settings page.
```suggestion
```

Tbaut (approved):
Looks good to me, just one line that shouldn't be needed.

packages/ui/cypress/tests/wallet-connect.cy.ts
asnaith: Ah yeah originally I was click to get through to the settings page but then changed that approach and forgot to take this out. Thanks

PR #465, opened 2023-12-12
Differenciate desktop and mobile to clean up tests

### LINKED ISSUE #447: Use unique element locators for multiproxy selector on desktop and mobile
At the moment we have to use **.first** when trying to interact with the multiproxySelector element eg: 

`topMenuItems.multiproxySelector().should('be.visible').first().click()`

This is because two elements with that locator can be found in the DOM (the desktop and mobile one). 

Let's give them unique names and update tests to get rid of the `.first`

Tbaut:
closes #447 

---

Submission checklist:

#### Layout

### DIFF
--- a/packages/ui/src/components/select/MultiProxySelection.tsx
+++ b/packages/ui/src/components/select/MultiProxySelection.tsx
@@ -13,7 +13,7 @@ import TextFieldLargeStyled from '../library/TextFieldLargeStyled'
 
 interface Props {
   className?: string
-  isDesktop: boolean
+  testId?: string
 }
 
 const getDisplayAddress = (option?: MultiProxy) =>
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/select/MultiProxySelection.tsx
```diff
@@ -27,7 +28,7 @@ const isOptionEqualToValue = (option?: MultiProxy, value?: MultiProxy) => {
   return option.multisigs[0].address === value.multisigs[0].address
 }
 
-const MultiProxySelection = ({ className }: Props) => {
+const MultiProxySelection = ({ className, isDesktop }: Props) => {
```
Lykhoyda, line 31: I would probably use the hook, to avoid passing the props to the component. F.e
```ts
const isDesktop = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'))
```

Lykhoyda (approved):
Looks good! There is a suggestion about isDesktop variable

packages/ui/src/components/select/MultiProxySelection.tsx
Tbaut: I don't think this hook should be used here. At the end of the day, it's just a `testId` to identify the specific `MultiProxySelection` component. The fact that we're using it for dektop or mobile is not relevant.

I'll rename it in `testId`.

asnaith (approved):
Looks great, thanks!

PR #466, opened 2023-12-12
Add more tests for create multisig

### LINKED ISSUE #445: More thorough multisig creation tests
After #436 that introduces a simple creation test, we should enhance it with checks
- that we can't continue with less than 2 signatories selected
- there's an error if the threshold is too high or low
- the name is optional
- there's an error message if no own account among the signatories (e.g just add random addresses)
- there's an error if not enough funds for the selected signer
- that creation without a pure works and shows the right text in the summary page (no mention of batch)
- that there's a remark created
- verify the behavior if we already have watched accounts with multisigs, e.g there's no big page with "Multisig creation in progress" but rather a banner "your new multisig is being created..."
- check that for a pure creation, there's a pure creation TX pending.

Lykhoyda:
closes #445 

---


packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,178 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+    })
+  })
+
+  describe('Multisig creation - step 3', () => {
+    beforeEach(() => {
+      topMenuItems.newMultisigButton().click()
+    })
+
+    it('Should show the error => if the funds are not sufficient to create Multisig', () => {
+      typeAndAdd(address4)
```
asnaith, line 166: Maybe we should call this variable "unfundedAccount" rather than "address4" so that it reads better?

packages/ui/src/components/NewMulisigAlert.tsx
```diff
@@ -14,7 +14,10 @@ const NewMulisigAlert = ({ className = '', onClose }: Props) => {
       variant="outlined"
       severity="info"
     >
-      <div className="infoText">
+      <div
+        className="infoText"
+        data-cy="info-multisig-created"
```
asnaith, line 19:
```suggestion
        data-cy="container-multisig-creation-info"
```

packages/ui/cypress/support/page-objects/landingPage.ts
```diff
@@ -13,7 +13,9 @@ export const landingPage = {
   noAccountFoundError: () => cy.get('[data-cy=label-no-account-found]', { timeout: 10000 }),
   linkedAddressNotFound: () => cy.get('[data-cy=label-linked-address-not-found]'),
   resetLinkedAddressButton: () => cy.get('[data-cy=button-reset-linked-address]'),
-  firstMultisigCreationLabel: () => cy.get('[data-cy=label-first-multisig-creation'),
+  firstMultisigCreationLabel: () =>
+    cy.get('[data-cy=label-first-multisig-creation', { timeout: 30000 }),
+  infoMultisigCreated: () => cy.get('[data-cy=info-multisig-created]'),
```
asnaith, line 18:
```suggestion
  infoMultisigCreated: () => cy.get('[data-cy=container-multisig-creation-info]'),
```

packages/ui/src/pages/Creation/Summary.tsx
```diff
@@ -150,7 +150,14 @@ const Summary = ({
           possibleSigners={possibleSigners}
         />
       </Box>
-      {!!errorMessage && !isSubmittingExtrinsic && <Alert severity="error">{errorMessage}</Alert>}
+      {!!errorMessage && !isSubmittingExtrinsic && (
+        <Alert
+          data-cy="error-not-enough-funds"
```
asnaith, line 155:
```suggestion
          data-cy="alert-insufficient-funds"
```

packages/ui/cypress/support/page-objects/newMultisigPage.ts
```diff
@@ -1,8 +1,21 @@
...
+    createMutlisigError: () => cy.get('[data-cy=create-multisig-error-message]')
+  },
+  step2: {
+    thresholdWarning: () => cy.get('[data-cy=input-warning-message]'),
+    thresholdInput: () => cy.get('[data-cy=input-threshold-creation'),
+    nameInput: () => cy.get('[data-cy=input-name-creation'),
+    checkboxUsePureProxy: () => cy.get('[data-cy=checkbox-use-pure-proxy]')
+  },
+  step3: {
+    errorNotEnoughFunds: () => cy.get('[data-cy=error-not-enough-funds]')
```
asnaith, line 19:
```suggestion
    errorNotEnoughFunds: () => cy.get('[data-cy=alert-insufficient-funds]')
```

packages/ui/src/pages/Creation/index.tsx
```diff
@@ -436,7 +436,14 @@ const MultisigCreation = ({ className }: Props) => {
         className="buttonContainer"
         flexDirection="column"
       >
-        {!!errorMessage && <div className="errorMessage">{errorMessage}</div>}
+        {!!errorMessage && (
+          <div
+            data-cy="create-multisig-error-message"
```
asnaith, line 441:
```suggestion
            data-cy="container-create-multisig-error"
```

asnaith (commented):
Left some minor suggestions but mainly on naming convention or typos. 

I think in this test context it makes sense to have the several describe blocks to logically separate each screen in the flow 👍

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,178 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+    it('Should not allow to click next => when creating a multisig with less than 2 Signatories selected', () => {
+      typeAndAdd(address1)
+
+      newMultisigPage.nextButton().should('be.disabled')
+    })
+
+    it('Should add signatory with no name given', () => {
+      typeAndAdd(randomSignatory1)
+
+      newMultisigPage.step1.accountNameInput().should('have.value', '')
```
Tbaut, line 113: maybe verify that we can add 2 like this and that the next button is enabled?

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,178 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+    })
+
+    it('should show the warning => when signatories address does NOT belong to your account', () => {
+      typeAndAdd(randomSignatory1)
+      typeAndAdd(randomSignatory2)
+
+      newMultisigPage.step1
+        .createMutlisigError()
+        .should('be.visible')
+        .and('contain', 'At least one of your own accounts must be a signatory.')
```
Tbaut, line 132: please verify that we can't go next, next button disabled.

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,178 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+    it('Should show the warning => when threshold is too high or low', () => {
+      typeAndAdd(address1)
+      typeAndAdd(address2)
+      newMultisigPage.nextButton().should('contain', 'Next').click()
+      const warningMessage = 'Threshold must be between'
+
+      // When is too high
+      newMultisigPage.step2.thresholdInput().type('3')
+      newMultisigPage.step2.thresholdWarning().should('be.visible').and('contain', warningMessage)
+      // When is too low
```
Tbaut, line 150: please check the "next" button status each time, that it's disabled, then enabled with the "2"

Tbaut (changes requested):
Thanks for this.

I'd like to do many more checks such as the ones in the issue
> that there's a remark created
verify the behavior if we already have watched accounts with multisigs, e.g there's no big page with "Multisig creation in progress" but rather a banner "your new multisig is being created..."
check that for a pure creation, there's a pure creation TX pending.

But also things like some error happening when for instance we type in twice the same signatory.

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,178 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+    cy.approveTx(txRequests[0].id)
+    notifications.notificationWrapper(10000).should('have.length', 1)
+    notifications.loadingNotificationIcon().should('be.visible')
+    notifications.notificationWrapper().should('contain', 'broadcast')
+
+    notifications.successNotificationIcon(30000).should('be.visible')
+    landingPage.firstMultisigCreationLabel().should('not.exist')
+    notifications.notificationWrapper().should('contain', 'Tx in block')
+
+    const expectedMultisigAddress = 'D9b1mkwhCwyRMUQZLyyKPdVkiJfFCuyVuWr3EmYAV6ETXkX'
```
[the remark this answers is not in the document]
Lykhoyda: I'm not sure if it's super generic. We already have many multisig addresses to use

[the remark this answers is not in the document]
Tbaut: Agreed, I don't think this will be much re-used either, since it's a function of the signatories and the threshold. I think it's ok to keep it here.

packages/ui/cypress/support/page-objects/landingPage.ts
```diff
@@ -13,7 +13,9 @@ export const landingPage = {
   noAccountFoundError: () => cy.get('[data-cy=label-no-account-found]', { timeout: 10000 }),
   linkedAddressNotFound: () => cy.get('[data-cy=label-linked-address-not-found]'),
   resetLinkedAddressButton: () => cy.get('[data-cy=button-reset-linked-address]'),
-  firstMultisigCreationLabel: () => cy.get('[data-cy=label-first-multisig-creation'),
+  firstMultisigCreationLabel: () =>
+    cy.get('[data-cy=label-first-multisig-creation', { timeout: 30000 }),
+  infoMultisigCreated: () => cy.get('[data-cy=container-multisig-creation-info]'),
```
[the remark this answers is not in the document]
asnaith: @Tbaut This was my suggestion, usually we prefix the locators with the element type. In this instance it's a div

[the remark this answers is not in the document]
Tbaut: But do we name things containers when they actually aren't? Alert, label would be more appropriate I thought. But yeah not a big deal, it's naming

packages/ui/cypress/fixtures/testAccounts.ts
```diff
@@ -62,6 +62,13 @@ export const testAccounts = {
     type: 'sr25519',
     mnemonic: 'bottom drive obey lake curtain smoke basket hold race lonely fit walk//chopsticks/3'
   } as InjectedAccountWitMnemonic,
+  'Not Funded Account 4 Chopsticks Kusama': {
```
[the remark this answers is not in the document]
Lykhoyda: Can I keep the changes in chopsticks yaml?

Lykhoyda:
@Tbaut and @asnaith thank you for your feedback! Please check again the updated version

packages/ui/cypress/fixtures/testAccounts.ts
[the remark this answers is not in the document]
Tbaut: Wait this doesn't make sense. This account `GecgyHrgdLEqkx34Ca5FW45QZ2wh66h3xbzzzgnxLKg2n8R` has a huge balance on Kusama. But it doesn't correspond at all with the mnemonic you've put here. 'bottom drive obey lake curtain smoke basket hold race lonely fit walk//chopsticks/4'

What you should have done is:
- check what account comes out with that mnemonic, and derivation it's "FtumrS4fFxcUL3UW5mxJ5ispXS4ADpGo3bas5ouQKbjxVZG" with public key `0x92e26f4dab1f112580136d3f0287207900a7ed74d9d1211b4062110599bf2930`
- This account obviously has 0 free KSM, so no need to force anything.

Tbaut (changes requested):
Please correct the account and remove the chopsticks config. There are an infinity of empty accounts, no need to fake one.

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -93,6 +127,14 @@ describe('Multisig creation', () => {
       acceptMultisigCreation()
 
       landingPage.infoMultisigCreated().should('be.visible')
+      accountDisplay.nameLabel().should('contain.text', multisigName)
+      multisigPage
+        .transactionList()
+        .should('be.visible')
+        .within(() => {
+          multisigPage.pendingTransactionItem().should('have.length', 2)
```
Tbaut, line 135: Why? This is not expect at all.

Tbaut (changes requested):
Please re-read my comments. I feel like half of what I suggested was ignored, without comment.

packages/ui/cypress/tests/multisig-creation.cy.ts
Lykhoyda: @Tbaut That's what I got when creating a pure. It's probably a bug as we returned the pending transaction which should be associated with previous test case "create Pure". Did I miss something? 

Screenshot 2023-12-15 at 18 06 14

packages/ui/cypress/support/page-objects/landingPage.ts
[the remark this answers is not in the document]
asnaith: I usually just go with the element type without too much thought but yeah to name it as what element it is used as is fine too I guess.

packages/ui/cypress/tests/multisig-creation.cy.ts
Tbaut: if you create twice the same multisig, with the same account, once with a pure, once without, then this is expected, but this test is far from what a normal user would do. We should create 2 different multisigs, one with a pure, another one without, the 2 multisigs should be different.

Tbaut:
I added some checks code to:
- check that the text mentions there will be 1 remark or 1 batch depending if we're creating with or without proxy
- check that the waiting tx is the one expected (pure creation or remark)
- fix a test that actually checks that it works to add a signatory without name, and that the account is displayed without name
- check that the first landing page waiting screen is displayed when creating a multisig for the first time
- check that there's a banner that disappears when it's not the first creation.

Tbaut (approved):
lmk @asnaith if there's anything important that you think is missing. Anton fixed the main issue, and I've added all the test I mentioned in my comments or that where mentioned in the issue. I think this is a pretty deep coverage.

Ideally, we should now add not only the creation of the multisig, but also the pure, and verify that the UI updates showing the pure in the header once it's available. This is outside of the scope of this issue, but I think it could be added by just extending the test with the pure creation.

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,290 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+    beforeEach(() => {
+      topMenuItems.newMultisigButton().click()
+      // Step 1
+      typeAndAdd(address1)
+      typeAndAdd(address2)
+      typeAndAdd(address3)
+      newMultisigPage.nextButton().should('contain', 'Next').click()
+    })
+
+    it('Create a multisig tx with pure proxy', () => {
```
asnaith, line 90:
```suggestion
    it('Create a multisig with a pure proxy', () => {
```

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -5,54 +5,290 @@ import { multisigPage } from '../support/page-objects/multisigPage'
...
+      multisigPage
+        .transactionList()
+        .should('be.visible')
+        .within(() => {
+          multisigPage.pendingTransactionItem().should('have.length', 1)
+          multisigPage.pendingTransactionCallName().should('contain.text', 'proxy.createPure')
+        })
+    })
+
+    it('Create a multisig tx without pure proxy', () => {
```
asnaith, line 125:
```suggestion
    it('Create a multisig without a pure proxy', () => {
```

asnaith (approved):
@Tbaut Took some time to look through it again, I think we're all good. This is very thorough now :) 

Pushed a minor wording change on a couple of test names

PR #468, opened 2023-12-21
Change RPC URLs - Polkadot and Kusama

DrW3RK:
rpc.polkadot.io and kusama-rpc.polkadot.io should not be used by or linked to within any dapps, as we cannot guarantee response times or service level. Ref: https://forum.polkadot.network/t/scaling-down-of-parity-s-public-infrastructure/4697

Adding an alternate URL. Feel free to modify.

closes #

---

Submission checklist:

#### Layout

- [ ] Change inspected in the desktop web ui
- [ ] Change inspected in the mobile web ui

#### Compatibility

- [ ] Functionality of change validated with a connected account with multisig
- [ ] Looks good for multisig with proxy

### DIFF
--- a/packages/ui/src/constants.ts
+++ b/packages/ui/src/constants.ts
@@ -39,7 +39,7 @@ export const networkList = {
   polkadot: {
     chainId: 'polkadot',
     explorerNetworkName: 'polkadot',
-    rpcUrl: 'wss://rpc.polkadot.io',
+    rpcUrl: 'wss://polkadot.api.onfinality.io',
     wsGraphqlUrl: 'wss://squid.subsquid.io/multix-arrow/v/v2/graphql',
     httpGraphqlUrl: 'https://squid.subsquid.io/multix-arrow/v/v2/graphql',
     logo: chainsPolkadotCircleSVG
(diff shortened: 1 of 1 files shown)


Tbaut:
Thanks a lot Radha. I knew it was being pulled out of pjs/apps, but I didn't know it was about to be decommissioned. How about the Rococo and Westend endpoints? I went ahead and used the ibp in https://github.com/ChainSafe/Multix/pull/469 everywhere we used the RPCs from Parity. Included the asset hubs.

DrW3RK:
Hey @Tbaut Yeah, I was experiencing issues with Multix for Polkadot and figured this might be the issue. "Parity will reduce the number of bootnodes we operate to 2 nodes per network, including the Polkadot and Kusama relay chains and their system parachains." I think the testnets are still under active support by Parity, so Rococo and Westend endpoints can remianed unchanged IMO. Thank you very much for being prompt!

Tbaut:
Got it, thanks Radha. When things are slow, it's generally high demand on the indexer side but we're taking care of this. Do not hesitate to ping me directly on any channel including telegram if you have issues.

PR #470, opened 2024-01-02
Use v2 of @chainsafe/cypress-polkadot-wallet

Tbaut:
no change beside naming of functions


packages/ui/cypress/support/Extension.ts
Tbaut, line 1: This should have been deleted before. It's not used any more.

asnaith (approved):

PR #472, opened 2024-01-08
Show watch indicator

### LINKED ISSUE #176: Differentiate between owned and watched accounts in drop down selector
When you watch an account eg ```5Fer9CC593nS2JTsS4SFC3iWT7M2cwcVWqixjAQ9uPDKg8Jg``` all of the pure's and multisigs that the address is behind will also appear as options in the drop-down selector. 

The issue here is that is confusing for a user to know which addresses belong to their own setup and which come from an account that is being watched. 

Screenshot 2023-05-24 at 11 15 00 AM

It would be good to differentiate in the list which addresses are being watched from the addresses that are part of the user's own setup. Perhaps with an eye or telescope emoji. 

Background: 

"There's no way to know what multisig/pure proxy a user want to see.
I think the confusion could be lowered if we could show clearly that the reason an account is in the dropdown is because of a watched account? The only way I would find easy to understand would be with :eyes: emoji next to the identicon or something :)"

_Originally posted by @Tbaut in https://github.com/ChainSafe/Multix/issues/161#issuecomment-1560182488_

Tbaut:
closes #176 

Not 100% convinced as this kinds of hides a good part of the identicon.. but on the other hand, we're talking about watched accounts.. so it's not that important.

image
image

---

Submission checklist:

#### Layout

#### Compatibility

### DIFF
--- a/packages/ui/cypress/support/page-objects/components/accountDisplay.ts
+++ b/packages/ui/cypress/support/page-objects/components/accountDisplay.ts
@@ -3,5 +3,6 @@ export const accountDisplay = {
   pureBadge: () => cy.get('[data-cy=badge-pure]'),
   multisigBadge: () => cy.get('[data-cy=badge-multi]'),
   nameLabel: () => cy.get('[data-cy=label-account-name]'),
-  addressLabel: () => cy.get('[data-cy=label-account-address]')
+  addressLabel: () => cy.get('[data-cy=label-account-address]'),
+  watchedIcon: () => cy.get('[data-cy=icon-watched]')
 }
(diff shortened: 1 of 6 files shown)


asnaith:
Looking good mostly but I found a problem where the "eye" badge from a watched account disappears once you connect a wallet that is a member of a pure. 

https://github.com/ChainSafe/Multix/assets/3982665/7390c68f-2ba1-4885-8dad-ddcba5319829

Tbaut:
This is actually expected. The moment you connect an account that is a signatory, the multi/proxy is not considered as "view only", and that's when you get the "new transaction" button, the ability to "review" txs etc. The only issue I see here, is that the account in the watch list become irrelevant, but there's not much we can do.

asnaith (approved):
I wasn't connecting with the account that was being watched...but it was part of the same multisig, so this would be intentional behaviour. 

So, this looks good to go! Thanks for adding the tests too :)

PR #482, opened 2024-01-24
Hotfix for WalletConnect connection

Tbaut:
Although the events aren't formerly supported (account change and disconnect, to be supported in another PR see https://github.com/ChainSafe/Multix/issues/483) this will make sure that the connection is actually working and doesn't crash.

I updated [the wiki](https://github.com/ChainSafe/Multix/wiki/Connect-Dapps-to-Multix-using-WalletConnect-v2) with up to date images.


Tbaut:
Merging as this is a very limited scope, and a hotfix

PR #490, opened 2024-01-26
update: display the type of proxy on multisig view

### LINKED ISSUE #79: Show the type of each proxy
When we create a pure proxy, it's by default an "Any" proxy.
Now any account can create other types of proxies. This is also supported by Multix, but it's not displayed anywhere.

E.g you can watch a signatory of `DCZyhphXsRLcW84G9WmWEXtAA8DKGtVGSFZLJYty8Ajjyfa` on Kusama (this is the ChaosDAO pure proxy) like `CrbJuFZWjY3yft424EMTY9hdbWoU878DFs74v3a8nNDeKJD` and you will see that there are 2 multisigs behind the pure proxy, one of which has the type "Any", the other "Governance".

The information regarding the type of proxy is available on the indexer in the `ProxyAccount` type.
https://github.com/ChainSafe/Multix/blob/main/squid/schema.graphql#L32

bolajahmad:
closes #79 
Show the Proxy Type
---

Submission checklist:

#### Layout

#### Compatibility

- [x] Applicable elements hidden / disabled for watched multisigs / pure: Hide Proxy type when Multi only


packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -26,6 +27,7 @@ const MultisigView = () => {
       <MultisigList>
         {selectedMultiProxy &&
           selectedMultiProxy.multisigs.map((multisig) => {
+            console.log({ multisig })
```
Tbaut, line 30: stray console.log

packages/ui/src/utils/proxyTypeRendererUtils.ts
```diff
@@ -0,0 +1,19 @@
+import { ProxyType } from '../../types-and-hooks'
```
Tbaut, line 1: Please see my comment in the issue, this is not future proof and having a dynamic way is better
https://github.com/ChainSafe/Multix/issues/79#issuecomment-1911936995

packages/ui/cypress/support/page-objects/multisigPage.ts
```diff
@@ -13,6 +13,7 @@ export const multisigPage = {
   multisigDetailsContainer: () => cy.get('[data-cy=container-multisig-details]'),
   multisigAccountSummary: () => cy.get('[data-cy=container-multisig-account-summary]'),
   thresholdListItem: () => cy.get('[data-cy=list-item-threshold]'),
+  proxyTypeListItem: () => cy.get('[data-cy=list-item-proxy-type]]'),
```
Tbaut, line 16: This is not used so far. It'd be great to have a test for it. And verify that the proxy type is correctly displayed.
Maybe in https://github.com/ChainSafe/Multix/blob/79cf6240aaee7926ed4b28d5fd89e17838b9c27a/packages/ui/cypress/tests/watched-accounts.cy.ts ?

Tbaut (commented):
Thank you so much for the PR. I left a couple comments

packages/ui/cypress/support/page-objects/multisigPage.ts
asnaith: Yeah it would be good to have a test for this. 

If @bolajahmad is not too familiar with cypress I can add one retroactively let's just make sure to open an issue if that's the case.

bolajahmad: Oh yes, please open an issue for it, there are also no tests for some related methods, like `thresholdListItem`. I will update this pull request soon.

Tbaut:
Can you also sign the CLA as per https://github.com/ChainSafe/Multix/pull/490#issuecomment-1911832032

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -134,6 +141,12 @@ const ListFieldText = styled('div')`
   color: ${({ theme }) => theme.custom.gray[800]};
 `
 
+const ListFieldValue = styled(Typography)`
+  color: ${({ theme }) => theme.custom.gray[800]};
```
Tbaut, line 145: The other values don't have the same color, it's `theme.custom.text.secondary`
```suggestion
  color: ${({ theme }) => theme.custom.text.secondary};
```

packages/ui/src/utils/stringUtils.ts
```diff
@@ -0,0 +1,4 @@
+export const camelcasetoString = (proxy: string) =>
```
Tbaut, line 1: can you change the name of the function and file to be `camelcaseToString` and `camelcaseToString.ts` and add it to the index.ts here https://github.com/ChainSafe/Multix/blob/1a82bd8cbc2ffefed5bb087ee8cf32afd54c4c1f/packages/ui/src/utils/index.ts ? The name of the file was too generic IMHO.

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -134,6 +141,12 @@ const ListFieldText = styled('div')`
   color: ${({ theme }) => theme.custom.gray[800]};
 `
 
+const ListFieldValue = styled(Typography)`
```
Tbaut, line 144: For consistancy, let's not use Typography, as we generally don't use it in the app.
```suggestion
const ListFieldValue = styled('div')`
```

And then you can remove the `as="p"` and the Typography import.

packages/ui/cypress/support/page-objects/multisigPage.ts
Tbaut: I created the issue, it shouldn't be a hard one https://github.com/ChainSafe/Multix/issues/494

Tbaut (commented):
It's looking great and working well. Tests are passing locally, I created an issue to add the tests. I've left a couple small nit that would be great to change, we can merge right after.

https://github.com/ChainSafe/Multix/issues/494

packages/ui/src/utils/stringUtils.ts
bolajahmad: It's definitely too generic. If another function comes related to string manupulation though, it would be cleaner to put them in a single stringUtils file.

I renamed the file in a new commit #[840f3eb](https://github.com/ChainSafe/Multix/pull/490/commits/840f3ebef7bf61e4c3df0dd30e06f49b59b70d78)

packages/ui/src/pages/Home/MultisigView.tsx
```diff
@@ -2,10 +2,11 @@ import AccountDisplay from '../../components/AccountDisplay'
 import { AccountBadge } from '../../types'
 // import MultisigActionMenu from './MultisigActionMenu'
 import { styled } from '@mui/material/styles'
-import { Chip } from '@mui/material'
+import { Chip, Typography } from '@mui/material'
```
Tbaut, line 5:
```suggestion
import { Chip } from '@mui/material'
```

Tbaut (approved):
Thank you so much :pray: :rocket:

asnaith:
Yeah! Thank you for this contribution @bolajahmad :)

PR #492, opened 2024-01-26
Automatically switch to polkadot when unsuported network

### LINKED ISSUE #486: Stuck on initializing when wrong network
- Type a wrong network in the url, e.g https://multix.chainsafe.io/?network=assethub-dot
- See an infinite loader "initialization"
- There's a console error but nothing else

--> need a proper error there, or an auto-redirect, or something

Tbaut:
closes #486

Default network is now Polkadot instead of Kusama.

We now redirect automatically to Polkadot when the selected network isn't supported instead of being stuck.
We also display the network in some error messages to show users in a clearer manner that the error is network related. As in "we found no account on Interlay", implying that they may have some e.g on Polkadot.

A test was added for the aut0-redirect.


asnaith (approved):
Makes sense to have the default Polakdot :) 

All looking good with this. I manually checked it out more as a formality and great we have the tests for it going forward 🚀

PR #496, opened 2024-01-29
Allow tests to run for external PRs

### LINKED ISSUE #495: Tests are not running for external PRs
This is annoying but is security sensitive because of the github secrets and push permissions to the prod instance of Multix

Tbaut:
closes #495 

This removes the walletconnect id, that is exposed in the UI in any case. Tests should run now at least for external contributors. Preview will still fail though.


.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# @tbaut owns any files in the /.github/
+
+# directory at the root of the repository and any of its
+
+# subdirectories.
+
+/.github/ @tbaut
```
Tbaut, line 7: Note that this shouldn't change much in terms of security. I did setup this repo to enforce my approval for any change in the .github dir, but there are probably other ways to exfiltrate our secrets than to just change these.

asnaith (approved):
Ah nice, yeah this makes sense - ID was discoverable via running the UI test anyway 👍

Issue #499, opened 2024-01-31
Better check for identity setting


Tbaut:
Check that the submitter of the identity (Multisig or proxy) has the required deposit
- show an error if not
- enhance https://github.com/ChainSafe/Multix/wiki/Why-are-funds-reserved%3F with the identity check

PR #502, opened 2024-02-02
Calculate identity deposit for the submitter

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/499

On Rococo the deposit depends on the bytes stored (how much space the the combined fields take), on Kusama/Polkadot it seems to be the amount of fields. In any case this is taken into account. When setting an identity, the account that makes the call is the one putting the deposit, if it's for a pure, it's the pure, if for the multisig, it's..  the multisig you got it.

Easy way to test is to have a multisig (with funds) and a pure without funds, and check when you select the one without funds that the UI isn't happy. In the test I went even further and just funded a multisig with exactly what's needed for a certain amonut of bytes. Try to set an identity with a larger name, and the error is displayed.

This may not fly in the long term, but we can change it then. The identity will move from the chains to a common-good chain, we'll have to use XCM to set it, and call the chain state to check it anyways.

---

Submission checklist:

#### Layout

#### Compatibility


asnaith (approved):
Nice fix with associated test too 😍

🚀🚀🚀

PR #506, opened 2024-02-27
Use indexer v3 with paseo reset

Tbaut:


Tbaut:
Tests are failing because Rococo is down.

PR #512, opened 2024-03-25
Add og image, desc etc for nice presentation in social media

Tbaut:
Preview of our links were not nice, since they didn't have an image. 
Before, e.g on Element
image

After on different platforms
image


Tbaut:
using
OG_IMAGE_BLUE_600px

Issue #305, opened 2023-08-18
Allow name edition per account


Tbaut:
Based on the conversation we had here: https://github.com/ChainSafe/Multix/issues/295

We should have a default name "Untitled". For the Pure, Multisig and Signatories on the home. The name should be styled lightly (maybe lighter than the current design) when it's untilted.
![image](https://github.com/ChainSafe/Multix/assets/33178835/69c022ae-20cc-4292-b36e-b5b9c357772d)

When clicking on the pencil, an input should show and allow the renaming of this specific account. Could you provide a design for this @sweetpea22 maybe? What would the buttons look like? Probably super small round icons?

We can keep the batch renaming modal, I think it's still useful.

PR #513, opened 2024-03-26
Allow easy name edition

Tbaut:
closes #https://github.com/ChainSafe/Multix/issues/305

I ended up changing the icon because I found the single pencil as in the screenshot from the issue not clear enough when small. This one isn't as nice, but it's very clear IMHO.

---

Submission checklist:

#### Layout

#### Compatibility


asnaith:
Fwiw I like the icon you went with more 😄

This is looking good for the most part but the implementation of it is a little weird on this screen, we only need one name field.

CleanShot 2024-03-27 at 10 20 30@2x

Tbaut:
Thanks for the review. You can't simply change the name though if I remove the ability here. You would have to delete, then add the same account again with a name this time.

asnaith (approved):
@Tbaut ok If you're good with it then, just wanted to bring it to your attention. I see your point though

Tbaut:
I originally didn't have it, and thought I'd it saves a couple clicks, I guess it doesn't hurt 🤷

Issue #489, opened 2024-01-25
Better way to copy an address, it's not clear right now


Tbaut:
When the address is displayed, or even with the identicon, it's not clear at all that the address will be copied.
We need UI indicators for this:
- maybe a tooltip above to say "click to copy address"
- maybe an overlay on hover, with the copy icon..

asnaith:
Tooltip or hover would be ok for desktop but perhaps not for mobile? 

I like the copy icon, as more or less anyone who has used a dapp before knows what this is for

![CleanShot 2024-01-29 at 12 02 06@2x](https://github.com/ChainSafe/Multix/assets/3982665/d26318b5-999c-4882-840e-ff1cf2fa1bfa)

We could then perhaps just display a label for a few seconds to say it's been copied?

asnaith:
Hmm actually with the number of identicons that we display that would be a lot of additional icons if we use a copy icon for each...nevermind 😅

Tbaut:
I think a click/tap on the address with a tooltip saying "address copied" would be better. This is really more for desktop for sure.. And not ideal for mobile.

PR #514, opened 2024-03-27
Enable copy address on click

Tbaut:
closes https://github.com/ChainSafe/Multix/issues/489

I went for the hover + copy tooltip
- so as to avoid having yet another icon (on top of the edit one)
- because our primary target is desktop and this is better than nothing, also mobile users may also figure it out (the tooltip will help)

The edge-cases here are in the dropdown menus where this function shouldn't be enabled.
I'd appreciate a check in browser I don't have access to such as Safari @asnaith and on mobile :pray

https://github.com/ChainSafe/Multix/assets/33178835/e209f8c4-8b92-46be-b69d-95d091ce744d

---

Submission checklist:

#### Layout

#### Compatibility


asnaith (approved):
Working as intended and looking good on macOS Safari, iOS Safari and Android Chrome

Desktop Safari

iOS Safari

Android Chrome

PR #516, opened 2024-03-27
Info for name override

### LINKED ISSUE #510: When a multisig has already been named on another network, we don't display the name and rather override it
**Description**
I create a multisig M1 on Polkadot, give it a name "M1-name"
I create a multisig on Kusama, with the same signatories, the process asks me for a name, it's "M2-name".

--> we've effectively overriden M1-name with M2-name without our user knowing.

**Solution**
At multisig creation, we should check whether the public key was already stored and if there's a name for the multisig. If there is, we should proceed and pre-fill the field with this name.

Tbaut:
closes #510

image

### DIFF
--- a/packages/ui/src/pages/Creation/NameSelection.tsx
+++ b/packages/ui/src/pages/Creation/NameSelection.tsx
@@ -1,27 +1,20 @@
-import { Box } from '@mui/material'
-import React, { useCallback, useState } from 'react'
+import { Alert, Box, styled } from '@mui/material'
+import React, { useCallback } from 'react'
 import { TextField } from '../../components/library'
 
 interface Props {
   className?: string
   name?: string
   setName: React.Dispatch<React.SetStateAction<string>>
+  originalName?: string
 }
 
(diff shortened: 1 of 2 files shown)


asnaith (approved):
Works as intended on the create msig screen 👍

Should we create an additional issue to add the same alert on the modal in the screenshot below because editing this changes it in for areas too?

CleanShot 2024-03-27 at 14 57 48@2x

Tbaut:
This is different as this modal is for name editing. if there's a name already, then it'll be displayed. In this pr, we're fixing a problem that users were overriding eventually a name, but without knowing it.

asnaith:
> This is different as this modal is for name editing. if there's a name already, then it'll be displayed. In this pr, we're fixing a problem that users were overriding eventually a name, but without knowing it.

Yeah I understand, it's a different situation, in edit mode the user is more aware of making a change to something already in existence rather than being unaware they're overriding something.

My thinking was that there could be a small set of users (specifically using setups without a pure) that may deliberately try to use a different name, per network, for the same multisig. Hypothetical though.

PR #518, opened 2024-04-02
Fix sub being displayed and display identity name in multiproxy selector

Tbaut:
Fixing the following issues:
1- The sub identity is displayed (here "Chainsafe Validator 0"), even when we have a local name. That's weird. E.g check on Kusama: Fqcoa6z2T8QJkNWMr8M7LcVmPonv7wwARsvciiL7HyUnJc4
image
2- The main selector shows the address even when there's an onchain identity
image

some quality of life bonus. Even if you have a local name (displayed by default) you can search by identity name, and it will be correctly displayed.


asnaith (approved):
Compared the branch side by side with prod, all looks good 👍

Tbaut:
Ah I added many more tests that I forgot to push.
- name display with sub (this PR in particular)
- name edition with check for buttons and keyboard interactions (esc or enter)
- check for the multi-proxy selector changes

PR #519, opened 2024-04-02
Add animation to copy hover

Tbaut:
the copy hover now appears gradually over 250ms


asnaith:
Not sure why the tests are failing but it's taking longer than normal to load the multisigs judging by the screenshots

Tbaut:
It's the indexer being slow thses days

PR #520, opened 2024-04-02
Bump subsquid deps

Tbaut:
bump subsquid dependancies and chopsticks with kusama forking block


Tbaut:
Chopsticks isn't happy with the bump, checking with the team here https://github.com/AcalaNetwork/chopsticks/issues/718

PR #528, opened 2024-04-09
Various UI and test fixes

Tbaut:
- remove indexer logs
- remove Kilt indexer until someone says they want it
- fix copy tooltip not disappearing
- remove cy.clock that's useless in tests and close the banner manually instead
- silences a warning in the console

### DIFF
--- a/packages/ui/cypress/tests/address-bar.cy.ts
+++ b/packages/ui/cypress/tests/address-bar.cy.ts
@@ -263,6 +263,9 @@ describe('Account address in the address bar', () => {
 
     topMenuItems.desktopMenu().within(() => topMenuItems.networkSelector().click())
     topMenuItems.networkSelectorOption('kusama').click()
+    landingPage
+      .noMultisigFoundError()
+      .should('contain.text', 'No multisig found for your accounts or watched accounts on kusama.')
     cy.url().should('not.include', 'address=')
   })
 
--- a/packages/ui/cypress/tests/multisig-creation.cy.ts
+++ b/packages/ui/cypress/tests/multisig-creation.cy.ts
@@ -144,14 +144,10 @@ describe('Multisig creation', () => {
 
       // The banner should be there, and disapear within 30s
       landingPage.multisigCreationInfoBanner().should('be.visible')
+      landingPage.creationInfoBannerCloseButton().click()
 
-      cy.clock().then((clock) => {
-        // The banner should disapear after 30s, speed it up by 15s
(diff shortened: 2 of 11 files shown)


packages/ui/cypress/tests/address-bar.cy.ts
```diff
@@ -263,6 +263,9 @@ describe('Account address in the address bar', () => {
 
     topMenuItems.desktopMenu().within(() => topMenuItems.networkSelector().click())
     topMenuItems.networkSelectorOption('kusama').click()
+    landingPage
```
Tbaut, line 266: @asnaith this actually made it so that it'd have failed earlier, I tested by removing my fix. It's better than having to wait 2s on purpose.

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -144,14 +145,12 @@ describe('Multisig creation', () => {
 
       // The banner should be there, and disapear within 30s
       landingPage.multisigCreationInfoBanner().should('be.visible')
+      landingPage.creationInfoBannerCloseButton().click()
 
-      cy.clock().then((clock) => {
```
Tbaut, line 148: this wasn't working. Adding the cy.clock before the cy.visit actually breaks everything, somehow, as the api calls don't get fired. All in all removing the clock

packages/ui/cypress/utils/waitForMultisigLength.ts
```diff
@@ -0,0 +1,13 @@
+import { multisigPage } from '../support/page-objects/multisigPage'
+import { topMenuItems } from '../support/page-objects/topMenuItems'
+
+export const waitForMultisigLength = (length: number) =>
+  cy.waitUntil(() => {
+    multisigPage.accountHeader().click()
```
Tbaut, line 6: this is doing a click outside basically, so that if any dropdown is open, it's close it.

packages/ui/cypress/tests/address-bar.cy.ts
asnaith: Yep makes sense :)

packages/ui/cypress/utils/waitForMultisigLength.ts
asnaith: Sounds good, might be worth adding a small comment there about that so we remember why we are doing that.

asnaith (approved):

PR #529, opened 2024-04-11
Autorefresh for tx so that pure and new msigs apprear without a manual refresh

Tbaut:
After implementing polling, unlike subscriptions, we now have to poll (for 1min) the new multisigs and proxies, after certain transactions to make sure the UI auto-refreshes.

### DIFF
--- a/packages/ui/cypress/tests/multisig-creation.cy.ts
+++ b/packages/ui/cypress/tests/multisig-creation.cy.ts
@@ -7,6 +7,7 @@ import { notifications } from '../support/page-objects/notifications'
 import { topMenuItems } from '../support/page-objects/topMenuItems'
 import { landingPage } from '../support/page-objects/landingPage'
 import { waitForTxRequest } from '../utils/waitForTxRequests'
+import { waitForMultisigLength } from '../utils/waitForMultisigLength'
 
 const fundedAccount1 = testAccounts['Funded Account 1 Chopsticks Kusama']
 const fundedAccount2 = testAccounts['Funded Account 2 Chopsticks Kusama']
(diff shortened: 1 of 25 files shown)


Tbaut:
Somehow the even extrinsic success isn't fired.. I'll ask the chopsticks team.
```
Transaction status: Ready
useSigningCallback.tsx:22 Transaction status: Broadcast
useSigningCallback.tsx:22 Transaction status: Finalized
useSigningCallback.tsx:87 Finalized block hash 0xdfd2e17fe46defa98bfc6b6e438de0530b241075593bbfa5fe8fe69b48e9a26f
```

asnaith (changes requested):
I'm seeing some UI glitchiness (I've tried Chrome and Firefox) that persists even after a hard refresh. 

No errors are shown in the console when this happens. 

https://github.com/ChainSafe/Multix/assets/3982665/fda675de-9abd-4fee-857c-cea422d06db1

Tbaut:
This ias What happens in tests as well that I can't reproduce. I'll manage

Tbaut:
Thanks @asnaith it should be fixed now.

packages/ui/cypress/tests/multisig-creation.cy.ts
```diff
@@ -92,7 +92,7 @@ describe('Multisig creation', () => {
       newMultisigPage.nextButton().should('contain', 'Next').click()
     })
 
-    it('Create a multisig with a pure proxy', () => {
+    it.only('Create a multisig with a pure proxy', () => {
```
asnaith, line 95:
```suggestion
    it('Create a multisig with a pure proxy', () => {
```

asnaith (approved):
Yep, all looking good now. No need for any hard refreshes. 

I added a small commit to remove a `it.only`

Tbaut:
Ah thanks, there's still quite some cleanup to do in this PR, I left some log and comments.

PR #537, opened 2024-05-15
New landing page

Tbaut:


Tbaut:
@sweetpea22 I've isolated the page and components about the landing page in a dedicated dir. It looks eventually messy in the app structure, but I can change it later, my goal is to provide and environment that is easy for you to jump in.

so basically:
- you can build everything inside this `LandingComponents`
- to see the result in the app either:
  - you have no connected or watched account, then this should be the landing page you see right away.
  - you have a connected account already with multisigs, then go to "About" in the top menu. This will be the live behavior, so that if you use Multix already, we want to show what you care about right away. If you don't, then we tell you more about Multix while inviting you to connect an account.
- I've built a super simple `Hero` component, with I think a good amount of things that are interesting such as: 
  - the buttons to connect etc, so you don't have to figure it out
  - a basic design so that you see how it's done throughout the app, basically style component with pure css
  - some usage of some common variables.

Unless you have it already, I highly recommend you to use the [vscode-styled-components](https://marketplace.visualstudio.com/items?itemName=styled-components.vscode-styled-components) plugin to have autocompletion and code styling.
  
 You should see that:
  
image

sweetpea22:
perfect, thank you!

Tbaut:
will not happen any time soone

Issue #543, opened 2024-06-26
Tests aren't passing, particularly related to identities


Tbaut:
Identity moved to the the ppl pallet. There needs to be new config to go check it out, per chain.

PR #546, opened 2024-06-28
Fix: bump cypress plugin

Tbaut:
self explanatory


Tbaut:
now there's only the identity failing, linked to https://github.com/ChainSafe/Multix/issues/543

PR #548, opened 2024-07-10
Fix: fix identity for chain with a People Chain (Kusama, soon Polkadot)

Tbaut:
Addresses: https://github.com/ChainSafe/Multix/issues/543


Tbaut:
youhou it's green again!

PR #598, opened 2024-12-05
Bump squid deps and fix Westend v5 tx version

Tbaut:
Update all deps of subsquid


Tbaut:
The 2 tests failing with Chopsticks are due to an issue with Subsquid. I'm asking for support

PR #608, opened 2024-12-15
Support calls with no params

Tbaut:
Close an issue happening with calls with no params. E.g `staking.chill`

image


Tbaut:
2 failing tests are passing locally. Merging since it's urgent.

PR #611, opened 2024-12-17
Support hydra specific calls

Tbaut:
Closes a problem with invalid checksum


Tbaut:
The same usual test failing :| merging bc hotfix in prod

PR #631, opened 2025-01-28
Allow hidding accounts

### LINKED ISSUE #58: Add the ability to delete a multisig
When you connect your wallet to Multix you will see all the multisigs that the connected account is part of displayed, it's a good UX. 

However, there may be occasions when a user does not want this or has been added to a multisig accidentially or unwillingly so it would be good to also provide the option for the user to delete / remove.

Tbaut:
closes #58 

---

Submission checklist:

#### Layout

#### Compatibility


Tbaut:
Thanks @asnaith I need to add some tests with this "don't show again" etc, and merge after, but I'm happy with the result so far.

PR #647, opened 2025-02-18
Add WalletConnect as a connection option

### LINKED ISSUE #622: Support wallet connect as a connection method.
Now that we're using dot-connect for the account connection, we should add support for Wallet Connect.

Tbaut:
closes #622


Tbaut:
tests are passing locally, I need to fix them :roll_eyes:

PR #668, opened 2025-04-01
Add back Moonbeam and Moonriver

Tbaut:
- not building due to some type error with papi descriptors
- ids on the indexer use addresses instead of pubkeys, needs fixing


Tbaut:
won't happen, they still basically don't use the native multisigs

PR #679, opened 2025-04-30
Fix kusama rpcs

Tbaut:
- use Kusama RPCs that support archive queries
- do not show decoded Txs on the overview when the calldata is > 1KB
- show in the console which RPC we're using and the connection status


Tbaut:
Known reason for the multisig test failing. Merging

PR #684, opened 2025-05-24
Bump deps and metadata

Tbaut:
maintenance


Tbaut:
known error in the tests due to chopsticks :(

PR #694, opened 2025-06-27
Print hex if not ascii

### LINKED ISSUE #693: Bytes are always decoded as text which is wrong for many extrinsics
We should try to decode it, see if it makes sense otherwise display as hex

from @albertov19

Tbaut:
closes #693


Tbaut:
network selector glitch in the test is failing. Unrelated.

